PR #676, opened 2014-12-19
Only apply live updates during main query and fix dashboard live updates bug

### LINKED ISSUE #675: Live updates should behave differently when filtering results
Live updates is a cool feature that brings new records into view using the Heartbeat API.

screen shot 2014-12-19 at 12 10 36 pm

However, these live updates do not respect filtered results. So when you are filtering results using the dropdown menus or search, live updates can come rolling in and skew the results.

To fix this, we should do one of the following:

1. Only allow live updates to run when the screen is using the default main query.
2. Filter live updates so they respect the query being used at the time.

In the short term, this problem can be avoided by simply turning live updates off.

screen_shot_2014-12-19_at_12_04_24_pm

See [this forum post](https://wordpress.org/support/topic/wp-cron-entries-in-log?replies=4#post-6346820) for more details.

fjarrett:
Resolves #675 

@shadyvb please review


ui/js/dashboard.js
```diff
@@ -67,6 +64,24 @@ jQuery( function( $ ) {
 
 	});
 
+	// Regenerate zebra stripes on record rows
+	function regenerate_row_alt() {
```
fjarrett: Done in https://github.com/wp-stream/stream/commit/4b2407d

ui/js/live-updates.js
```diff
@@ -98,6 +106,24 @@ jQuery( function( $ ) {
 
 		});
 
+		// Regenerate zebra stripes on record rows
+		function regenerate_row_alt() {
```
fjarrett: Done in https://github.com/wp-stream/stream/commit/4b2407d

PR #684, opened 2015-01-14
Handle bool values appropriately during wp_stream_log_data filter

### LINKED ISSUE #680: PHP Catchable Fatal Error on EDD Checkout Page
WordPress SEO Connector is throwing PHP Catchable Fatal error on EDD checkout page. Hence Store is breaking.

screenshot from 2015-01-07 16 52 51

Not really sure if we should remove `array` keyword from parameters in this function [here](https://github.com/wp-stream/stream/blob/develop/connectors/class-wp-stream-connector-wordpress-seo.php#L348) or not.

Submitting a PR as of now. Open to other suggestions or discussion

fjarrett:
Fixes #680

@desaiuditd Please test this branch in your environment and let me know if the fatal is gone.

### DIFF
--- a/connectors/class-wp-stream-connector-acf.php
+++ b/connectors/class-wp-stream-connector-acf.php
@@ -453,7 +453,11 @@ class WP_Stream_Connector_ACF extends WP_Stream_Connector {
 	 *
 	 * @return array|bool
 	 */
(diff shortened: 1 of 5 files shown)


fjarrett:
@desaiuditd This is working for me, so I've merged it.

Issue #499, opened 2014-05-02
Introduce WP-CLI stream command


westonruter:
It would be really cool if Stream data could be queried via WP-CLI, and if settings could be updated.

For example, there should be a `wp stream log` command that takes the filter arguments like `--author=1` or `--context=sidebars`.

Also, a cool use case would be something like `wp stream log --tail` which would output a continual stream of entries as they come in. The PHP process would keep running until `Ctrl-C`. Basically like Heartbeat on the command line. You could then pipe the results into something else.

fjarrett:
I've created a [Query Reference](https://github.com/wp-stream/stream/wiki/Query-Reference) in our Wiki that can be used as a guide when using the WP-CLI command.

fjarrett:
Released in `2.0.3`

szepeviktor:
Could you please include the Query Ref. as PHP-doc?
E.g. https://github.com/wp-cli/wp-cli/blob/master/php/commands/post.php#L218-L278

fjarrett:
Hi @szepeviktor, thanks for the suggestion.

Would you be willing to providing a pull request for this?

We have [Query Parameters documented](https://github.com/wp-stream/stream/wiki/Query-Reference) in the wiki. Otherwise we will just add these to the docblock when we get around to it.

PR #685, opened 2015-01-15
Adding support for WP-CLI command

fjarrett:
Implements #499 

I'd like to get this out in it's most basic form first, and improve on it later down the line.

The scope of this PR does not address `tail` or `export` args which could certainly be added in the future.

Right now there is only a `query` subcommand and it will accept any associative argument(s) that the `WP_Stream_Query::instance()->query()` method accepts. You can add as many of these as you'd like in parallel.

**Some examples:**

```
wp stream query
wp stream query --author=1
wp stream query --author=1 --records_per_page=20
wp stream query --author=1 --action=login --date_before=2015-01-15 --order=asc
wp stream query --author_role__not_in=administrator,editor
```

If you don't like the output format you could also specify which fields you want to output to the console using the `--fields` arg, adding multiple fields in a comma separated list.

**Some examples:**

```
wp stream query --author=1 --fields=summary
wp stream query --author=1 --fields=created,ip,summary
wp stream query --author_role=editor --date_before=2015-01-01T12:00:00 --fields=created,context,action,summary
```

The fields for default output are also filterable, so you can do:

```php
add_filter( 'wp_stream_wp_cli_default_fields', function() {
	return array( 'created', 'author_meta.display_name', 'context', 'action' );
});
```

**Update 2015-01-16**

I have added two new subcommands that accept no arguments: `wp stream account` and `wp stream help`

### DIFF
--- a/includes/wp-cli.php
+++ b/includes/wp-cli.php
@@ -85,68 +85,12 @@ class WP_Stream_WP_CLI_Command extends WP_CLI_Command {
 	}
 
 	/**
-	 * Subcommand to retreive details about your Stream account
-	 *
-	 * @subcommand account
-	 */
-	public function account() {
-		self::connection();
-
-		$site       = WP_Stream::$api->get_site();
-		$plan       = isset( $site->plan->type ) ? $site->plan->type : 'unknown';
-		$plan_label = ( 'free' === $plan ) ? __( 'Free', 'stream' ) : ( 'pro_monthly' === $plan ) ? __( 'Pro', 'stream' ) : ucwords( $plan );
-
-		WP_CLI::line( sprintf( __( 'Plan: %s', 'stream' ), $plan_label ) );
-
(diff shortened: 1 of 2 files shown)


fjarrett:
@westonruter since you were the OP for this feature, mind giving it a go?

westonruter:
Love to, but I'm not going to have time in the foreseeable future.

fjarrett:
@westonruter No worries, thanks man!

fjarrett:
The scope of this PR does not address `tail` or `export` args which could certainly be added in the future.

fjarrett:
I decided to remove the `account` and `help` subcommands for now to keep with my original goal of keeping this feature simple to start and expanding on it later.

I think the `query` subcommand and already robust enough for people to benefit from.

fjarrett:
I've created a [Query Reference](https://github.com/wp-stream/stream/wiki/Query-Reference) in our Wiki that can be used as a reference when using this new WP-CLI command.

PR #687, opened 2015-01-16
Hook load methods on init rather than plugins_loaded

### LINKED ISSUE #686: PHP notice on BuddyPress sites for bp_setup_current_user
A debug backtrace shows that Stream is causing this error by loading `WP_Stream_Admin::load()` too early on `plugins_loaded`.

It should be loaded on `init` after the `$current_user` is properly set up. And we may as well be hooking into the load methods for other classes on `init` as well for good measure.

```
Notice: bp_setup_current_user was called incorrectly.
The current user is being initialized without using $wp->init().
Please see Debugging in WordPress for more information.
This message was added in version 1.7 in /srv/www/wordpress-trunk/wp-includes/functions.php on line 3550
```

See https://buddypress.trac.wordpress.org/ticket/4830

fjarrett:
Fixes #686

### DIFF
--- a/extensions/notifications/class-wp-stream-notifications.php
+++ b/extensions/notifications/class-wp-stream-notifications.php
@@ -75,7 +75,7 @@ class WP_Stream_Notifications {
 			return;
 		}
 
-		add_action( 'plugins_loaded', array( $this, 'load' ) );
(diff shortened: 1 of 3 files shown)


fjarrett:
This is working fine for me, so I'm going to merge it.

PR #690, opened 2015-01-20
Order searches by date desc by default, instead of by _score

### LINKED ISSUE #689: Search results ordered by relevance
Currently when searching Stream records, results are ordered by relevance.

For example, if I do a search for `Stream` I get:
screen shot 2015-01-20 at 11 05 25

In my opinion, it makes more sense to order them by date. Stream records are always presented by date except in this case and it confuses me.

fjarrett:
Resolves #689

@lukecarbis Please test this out.


fjarrett:
This was a simple one-liner, and it's working for me.

PR #691, opened 2015-01-20
Fix issues when N/A user is selected in exclude rules

### LINKED ISSUE #688: Can't exclude N/A user
Firstly, when adding the user "N/A" to the exclude rules and pressing save, the page reloads with nothing selected in the Author or Role dropdown.

screen shot 2015-01-20 at 10 56 28

Secondly, N/A entries are being tracked, despite the (broken) exclude rule.

screen shot 2015-01-20 at 10 56 38

fjarrett:
Fixes #688 

@lukecarbis Please test this out. It should fix both the name issue not displaying when selected in a rule as well as the issue with the rule not being enforced.


fjarrett:
This is working for me.

PR #692, opened 2015-01-20
Disable logging during the content import process

### LINKED ISSUE #672: Silently disable Stream during content import
A user in the WordPress forums [pointed out](https://wordpress.org/support/topic/there-was-a-problem-verifying-your-site-with-stream-please-try-again-later?replies=1) that deactivating Stream requires reconnecting upon reactivation but that it can be frustrating when wanting to import content on a regular basis because we prompt the user to deactivate Stream before allowing importing due to the mass amounts of superfluous activity records that result.

screen shot 2014-12-12 at 9 04 53 pm

It sounds like there is an opportunity for us to improve the experience by disabling Stream silently during an import process rather then requiring that the plugin be deactivated. We need to investigate core and see such detection is possible.

fjarrett:
Resolves #672

### DIFF
--- a/classes/class-wp-stream-admin.php
+++ b/classes/class-wp-stream-admin.php
@@ -430,7 +430,6 @@ class WP_Stream_Admin {
 				'bulk_actions' => array(
 					'i18n' => array(
 						'confirm_action' => sprintf( __( 'Are you sure you want to perform bulk actions on over %s items? This process could take a while to complete.', 'stream' ), number_format( absint( $bulk_actions_threshold ) ) ),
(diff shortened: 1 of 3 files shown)


fjarrett:
@jahrl Since you were the [OP for this feature](https://wordpress.org/support/topic/there-was-a-problem-verifying-your-site-with-stream-please-try-again-later?replies=1) would you be able/willing to test out this branch?

https://github.com/wp-stream/stream/archive/issue-672.zip

fjarrett:
I'm merging this because it's working nicely for me and my discovery of the `WP_IMPORTING` constant made the code needed for this feature quite lean.

PR #696, opened 2015-01-21
Get plugins to be deleted from request rather than global var

### LINKED ISSUE #695: PHP warning when deleting plugins
A [user reported](https://wordpress.org/support/topic/issues-deleting-other-plugins?replies=1) in the plugin forum that they are receiving an error when attempting to delete a plugin, and it's preventing the plugin from being removed.

```
Warning: Invalid argument supplied for foreach() in wp-content/plugins/stream/connectors/class-wp-stream-connector-installer.php on line 309
```

Not only that, but during my testing of this issue I've discovered that Stream is not tracking plugin deletions at all when done in bulk. So that will need to be addressed as well.

fjarrett:
Fixes #695


fjarrett:
This is working great for me.

PR #699, opened 2015-02-03
Allow property values of 0 to be recognized as valid

### LINKED ISSUE #698: Allow properties with zero values to be included in queries
We are currently disallowing properties with [empty values](http://php.net/manual/en/function.empty.php) from being included in the final query array.

This is a problem if you want to query something with an ID of zero, such as a `user_id`.

We should adjust our definition of what an "empty" property is when building the final query.

**Disallowed values**

* `NULL`
* `bool(false)`
* `string(0) ""` (an empty string)
* `array(0) {}` (an empty array)

**Allowed values**
* `string(1) "0"`
* `int(0)`

fjarrett:
Fixes #698

### DIFF
--- a/classes/class-wp-stream-query.php
+++ b/classes/class-wp-stream-query.php
@@ -166,7 +166,7 @@ class WP_Stream_Query {
 
 		// PARSE PROPERTIES
 		foreach ( $properties as $property => $default ) {
-			if ( $args[ $property ] ) {
+			if ( isset( $args[ $property ] ) && '' !== $args[ $property ] && false !== $args[ $property ] && ! is_array( $args[ $property ] ) ) {
 				$filters[]['term'][ $property ] = $args[ $property ];
 			}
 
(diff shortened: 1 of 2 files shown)


classes/class-wp-stream-query.php
```diff
@@ -166,7 +166,7 @@ public function query( $args ) {
 
 		// PARSE PROPERTIES
 		foreach ( $properties as $property => $default ) {
-			if ( $args[ $property ] ) {
+			if ( isset( $args[ $property ] ) && '' !== $args[ $property ] && false !== $args[ $property ] && ! is_array( $args[ $property ] ) ) {
```
fjarrett: I've also opted to totally ignore array values altogether since they should never be allowed in single property values, only `__in` and `__not_in` property variants should have array values.

fjarrett:
It turns out that this fix is also somewhat related to #688 which pointed out some issues happening with the N/A user, which has a `user_id` of `0`.

PR #707, opened 2015-03-02
Add global hooks even if user is excluded

jonathanbardo:
Right now if the logging process is disabled for a particular user in Stream settings the user will not see any actions links on Stream list page because connectors are not loaded for him.

screen shot 2015-03-02 at 11 18 13 am

------

screen_shot_2015-03-02_at_11_22_30_am

I propose adding a a global function that will load actions and filters from connectors that need to be loaded even if the logging is disabled for a particular user.

### DIFF
--- a/classes/class-wp-stream-connectors.php
+++ b/classes/class-wp-stream-connectors.php
@@ -124,7 +124,7 @@ class WP_Stream_Connectors {
 			$is_excluded_connector = apply_filters( 'wp_stream_check_connector_is_excluded', $is_excluded, $connector_name, $excluded_connectors );
 
 			$connector::register_global_hooks();
-			
+
 			if ( $is_excluded_connector ) {
 				continue;
 			}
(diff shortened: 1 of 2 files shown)


classes/class-wp-stream-connectors.php
```diff
@@ -123,6 +123,8 @@ public static function load() {
 			 */
 			$is_excluded_connector = apply_filters( 'wp_stream_check_connector_is_excluded', $is_excluded, $connector_name, $excluded_connectors );
 
+			$connector::register_global_hooks();
+
```
lukecarbis: Can you please remove the whitespace here?

lukecarbis:
I love the concept here, @jonathanbardo. Could you update your PR to include implementation inside the list table?

jonathanbardo:
This should already work inside the list table. I'm not sure what I'm missing here...

fjarrett:
@jonathanbardo It looks like you are using an old version of Stream `1.4.x`. I am not able to reproduce the issue in `2.x`.

PR #723, opened 2015-05-15
Updated Unit Tests #611

### LINKED ISSUE #611: Updated Unit Tests
Our unit tests are written primarily for the currently released version of Stream. They should be updated for Stream 2.0 to include things like the extensions and API.

solace:
* Added support for querying `author_meta`.
* changed `$post->guid to get_attached_file($post->ID)` to get the actual file for `attachment_types`
* `callback_profile_update` log should use updated information
* Added unit tests for comments, media, menus, posts, settings, taxonomy, users and widgets
* Added unit test configuration settings

Resolves #611


fjarrett:
@solace The `WP_Stream_Connector_Comments` tests seem to be having trouble passing.

See https://travis-ci.org/wp-stream/stream/jobs/62732449#L933

solace:
@fjarrett A couple things:

1. In order to test this, it needs to have an API key and SITE_UUID passed in to be able to write records.  I run it on my local with `SITE_UUID=site-770c1180-dcd2-11e4-a38c-b580111fac68 API_KEY=jgsopafdiplgf9tg4tudzxltfrhianudyaqefv3stqnzb6zt phpunit --verbose --debug`  I'm not sure how we're to address this using travis.

2. Side note, while the unit tests creates and destroys WP data, these tests will add about 100-150 records each time they're run. Because these records don't ever get removed from ES, so the tests create records with timestamps somewhere in the meta so that they can be uniquely identified. In the cases where there is no way to do this, we do a query before we expect the write, and after to ensure the number of matching records increases by 1.

PR #726, opened 2015-07-01
Namespacification and Antistatication

lukecarbis:
WIP!

In it's current form, this PR will break everything. So no touching, just looking, while I continue work on it.

It's basically a rewrite of Stream, in case you were wondering.

Getting rid of all the silly static workaround we had in place, and namespacing everything.

Here's what's left to do:

- [x] Destaticise connectors
- [x] Modify WP CLI integration to be more integrated with the plugin
- [ ] Try to remove all functions in the functions.php file, or at least namespace them
- [x] Test to see if we still need the Records class
- [x] Search and remove all unneeded instances of the terms `static`, `self`, and `__CLASS__`
- [x] Possibly uses the `get_stream_instance()` helper function inside the `Author` class?
- [x] Pull out Dashboard Widget
- [x] Pull out Unread
- [ ] MERGE LATEST FROM LITE!!

Oh, and by the way, this PR pulls out Extensions and Reports entirely. :) You're welcome (I think?).


stream.php
```diff
@@ -29,361 +29,23 @@
...
-		return compact( 'dir_url', 'dir_path', 'dir_basename', 'plugin_basename' );
-	}
-
-}
-
-if ( ! WP_Stream::is_valid_php_version() ) {
-	WP_Stream::fail_php_version();
+if ( ! version_compare( PHP_VERSION, '5.3', '>=' ) ) {
+	load_plugin_textdomain( 'stream', false, dirname( plugin_basename( __FILE__ ) ) . '/languages/' );
+	add_action( 'shutdown', function() {
```
fjarrett: @lukecarbis You can't use an anonymous function here because that would require PHP 5.3 :wink:

fjarrett:
@lukecarbis I'm getting a fatal error after the merge.

```
Fatal error: Call to a member function get_user_feed_settings_admin_url() on a non-object in /wp-content/plugins/stream/classes/class-settings.php on line 314
```

Looks like this is because the `$settings` property is called before the `$feeds` property in `\WP_Stream\Plugin` so it's not able to reference the object yet. However, the `\WP_Stream\Feeds`  object also references settings, so we have a chicken egg problem here.

I'm not sure how I feel about `\WP_Stream\Plugin`, as it is right now it's pretty much a God object.

lukecarbis:
I'll look into that shortly. I agree that `Plugin` is a bit of a God object right now (`Admin`, too). This can be remedied by moving some of it's functions to more helpful places.

`Plugin` should be used only for high-level plugin functionality methods, like `i18n`, `autoload`, and preparing other classes for use.

`is_vip()` should be in `functions.php`
`is_wp_cron_enabled()` should be in `functions.php`
`disable_logging_during_import()` should be in `Admin`
`notice()` should be in `Admin`
`admin_notices()` should be in `Admin`
`verify_db()` should be in `Install`
`install()` can stay where it is
`update_activation_hook()` should be in `Install`
`register_update_hook()` should be in `Install`
`frontend_indicator()` should stay where it is, or be put into it's own `Frontend` class
`locate_plugin()` can stay
`get_version()` can stay

lukecarbis:
Fatal error is gone now.

lukecarbis:
@fjarrett I've made all those changes now. :)

PR #735, opened 2015-08-25
Stream 3: Fix \WP_Stream\List_Table namespace so live updates work correctly

rob:
On a fresh Stream 3 install, live updates fail silently with the following error:

    PHP Fatal error:  Class 'WP_Stream\\WP_Stream_List_Table' not found in /[...]/wp-content/plugins/stream/classes/class-live-update.php on line 179

Due to PHP's namespaces, this needs to be `\WP_Stream\List_Table` or just `List_Table`, not `WP_Stream_List_Table`. This commit fixes this so live updates work correctly again like they should.

### DIFF
--- a/classes/class-live-update.php
+++ b/classes/class-live-update.php
@@ -176,7 +176,7 @@ class Live_Update {
 		$enable_stream_update = ( 'off' !== $this->plugin->admin->get_user_meta( get_current_user_id(), $this->user_meta_key ) );
 
 		// Register list table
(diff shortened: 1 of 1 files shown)


lukecarbis:
Thanks so much @rob. Wow - how did you get _that_ github username!?

Your PR looks great, and I've tested it and it works great. Would you mind issuing it against the `develop` branch instead of `master`?

Thanks.

lukecarbis:
Also, I have no idea how I missed this.

rob:
OK @lukecarbis, I just issued it against the `develop` branch.

And I registered this username when GitHub was first announced, I think. It sucks when people use the shorthand @rob to reference other people and I end up getting notifications though when it's not related to me. :-)

PR #739, opened 2015-08-26
Stream 3: Fix \WP_Stream\List_Table namespace so live updates work correctly

rob:
*(Reposting from #735 to issue PR against `develop` branch instead of `master`)*

On a fresh Stream 3 install, live updates fail silently with the following error:

    PHP Fatal error:  Class 'WP_Stream\\WP_Stream_List_Table' not found in /[...]/wp-content/plugins/stream/classes/class-live-update.php on line 179

Due to PHP's namespaces, this needs to be `\WP_Stream\List_Table` or just `List_Table`, not `WP_Stream_List_Table`. This commit fixes this so live updates work correctly again like they should.


lukecarbis:
Legend. Thanks for this! :ok_hand:

Issue #501, opened 2014-05-03
Switching user is incorrectly tracked


kucrut:
Related to #434, when switching user using the [User Switching](https://wordpress.org/plugins/user-switching/) plugin, Stream only tracks the logging out part of the first user, when it should be tracking the logging in part of the second user. Also, when switching back, Stream doesn't log anything.

PR #744, opened 2015-08-28
Implement a User Switching connector

johnbillion:
This adds a new connector which integrates with the [User Switching](https://wordpress.org/plugins/user-switching/) plugin.

Stream works fine with User Switching (see #501), but it's a little sub-optimal. When you switch user, two entries get logged, one for the logout action and one for the login action.

This connector uses the actions that the User Switching plugin fires to more accurately log the actions performed (eg. `Switched user to editor (Editor)`), and also unhooks the default Users connector actions so the logout and login actions aren't logged. The connector also logs its actions under the Users context rather than its own.

New actions that are introduced:

 * Switched User ([`switch_to_user` action](https://github.com/johnbillion/user-switching/blob/1.0.8/user-switching.php#L888-L896))
 * Switched Back ([`switch_back_user` action](https://github.com/johnbillion/user-switching/blob/1.0.8/user-switching.php#L888-L896))
 * Switched Off ([`switch_off` action](https://github.com/johnbillion/user-switching/blob/1.0.8/user-switching.php#L888-L896))

--

I've had to make two small changes Stream's `Connector` and `Connectors` classes in addition to adding the new connector. Currently, it's impossible to unhook a connector's actions due to [the anonymous instance of each connector which gets instantiated inside itself](https://github.com/xwp/stream/blob/master/classes/class-connector.php#L37-L42). There's no need for this, so I've switched to using `$this`.

In addition, I've added `$this` as a parameter to the `wp_stream_after_connectors_registration` action so a connector can use it to unhook actions attached to any of its connectors. This is necessary because the `connectors` property on the global `$wp_stream` object hasn't been populated at this point.

### DIFF
--- a/classes/class-connector.php
+++ b/classes/class-connector.php
@@ -34,14 +34,11 @@ abstract class Connector {
 	 * Register all context hooks
 	 */
 	public function register() {
-		$class_name = get_called_class();
-		$class = new $class_name;
-
-		foreach ( $class->actions as $action ) {
-			add_action( $action, array( $class, 'callback' ), null, 5 );
+		foreach ( $this->actions as $action ) {
+			add_action( $action, array( $this, 'callback' ), 10, 5 );
 		}
 
-		add_filter( 'wp_stream_action_links_' . $class->name, array( $class, 'action_links' ), 10, 2 );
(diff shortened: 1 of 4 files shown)


lukecarbis:
This is great, @johnbillion. Thanks so much!

PR #777, opened 2015-10-02
Fix Gravity Forms error when deleting forms / track when form trashed

### LINKED ISSUE #771: Gravity forms error when deleting a form.
`( ! ) Strict standards: Only variables should be passed by reference in /srv/www/wordpress-develop/src/wp-content/plugins/stream/connectors/class-connector-gravityforms.php on line 726`

Also, trashing a form isn't tracked.

rob:
This fixes #771.

It also replaces a deprecated Gravity Forms `get_forms_by_id()` function with `get_form_meta_by_id()`.


rob:
Btw - just added the ability to track when the form is trashed, too. Should close #771 completely.

PR #779, opened 2015-10-06
Gravity forms connector update

lukecarbis:
This updates the Stream Gravity Forms connector to work with the renamed actions in Gravity Forms.

Needs testing:
- Deleting notification
- Deleting confirmation
- Resetting a form's views
- Deleting / Restoring a form
- Activating / Deactivating a form
- Duplicating a form
- Exporting a form / forms
- Exporting entries

@rob Could you please test and review these changes?

PR #780, opened 2015-10-07
Gravity Forms connector update - bugfix/issue-771

rob:
This PR extends #779 with some more fixes. (also see #771)

FYI: All of the current GF export/import tracking functionality is broken in Stream. There's no longer a way to track form exports (and this was causing a bug - see below), and the actions for exporting entries and importing forms has changed. This commit fixes export/import tracking.

I also bumped `PLUGIN_MIN_VERSION` due to the new `gform_forms_post_import` action introduced in `1.9.13.29` (latest stable release.)

I discovered a bug in the current version. The `gform_export_form` line (which is actually a filter, not action) was causing the Gravity Forms .json export file to be empty. After removing that line, exports are working fine.

### DIFF
--- a/connectors/class-connector-gravityforms.php
+++ b/connectors/class-connector-gravityforms.php
@@ -1,7 +1,9 @@
 <?php
+
 namespace WP_Stream;
 
 class Connector_GravityForms extends Connector {
+
 	/**
 	 * Connector slug
 	 *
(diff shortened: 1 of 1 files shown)


lukecarbis:
Hey @rob - where are we at with this?

rob:
Hi @lukecarbis - sorry, almost done (busy work week at my 'normal' job.) I will hopefully have everything committed tonight!

lukecarbis:
@rob Ready for me to test?

rob:
Hi @lukecarbis - sorry for the delays, my day job has been very busy lately. I did manage to do some work and clean up a lot of stuff, and it should work as it stands right now without any errors. But there's still some actions and other stuff I want to go over and clean up still, so it's your call.

Issue #778, opened 2015-10-05
PHP 5.2 gives fatal error


lukecarbis:
Stream should fail gracefully on PHP versions lower than 5.3, but currently it's not.

Getting this error:
```
Plugin could not be activated because it triggered a fatal error.

Parse error: syntax error, unexpected T_STRING in /home/content/79/7684179/html/el/wp-content/plugins/stream/stream.php on line 37
```

Reported here: https://wordpress.org/support/topic/fatal-error-on-activation-in-wp-431-stream-301?replies=3#post-7507734

PR #781, opened 2015-10-08
Stream fails gracefully on PHP versions lower than 5.3

greguly:
Fixes https://github.com/xwp/stream/issues/778


lukecarbis:
Thanks so much for this PR @greguly. Could you kindly revert c2703b9, and revert the plugin version to 3.0.2? I'll do that myself in the `develop` branch when I'm ready to release it.

Also - rather than including a new file, have you considered this approach? I feel it's a lot cleaner, but I've not tested if it works:

```
if ( ! version_compare( PHP_VERSION, '5.3', '>=' ) ) {
 	load_plugin_textdomain( 'stream', false, dirname( plugin_basename( __FILE__ ) ) . '/languages/' );
 	add_action( 'shutdown', 'wp_stream_fail_php_version' );
 } else {
	require __DIR__ . '/classes/class-plugin.php';
	$plugin_class_name = 'WP_Stream\\Plugin';
	if ( class_exists( $plugin_class_name ) ) {
		$GLOBALS['wp_stream'] = new $plugin_class_name();
	}
 }
```

Finally, what's your rationale for removing `wp_stream_get_instance()`? I know a lot of Stream extensions use this as a cleaner / future-proof way of retrieving the Stream instance.

rob:
@lukecarbis, I can confirm your code snippet works with PHP 5.2. I like your approach better instead of including another file. (I don't think you need to escape the backslash in `WP_Stream\\Plugin` though in this case.)

greguly:
@rob  got it correctly, there is no need to escape the backslash :-)

Of course will revert https://github.com/xwp/stream/commit/c2703b9286173ba8450f90017696efad3c24a22b, will also fix the backslash and use your solution.

My rationale for including a new file was to isolate all 5.3 code in one place, if you pay attention function wp_stream_get_instance() was moved there too.

greguly:
All set :-)

stream.php
```diff
@@ -3,7 +3,7 @@
  * Plugin Name: Stream
  * Plugin URI: https://wp-stream.com/
  * Description: Stream tracks logged-in user activity so you can monitor every change made on your WordPress site in beautifully organized detail. All activity is organized by context, action and IP address for easy filtering. Developers can extend Stream with custom connectors to log any kind of action.
- * Version: 3.0.2
+ * Version: 3.0.3
```
lukecarbis: @greguly Can you please just switch this line back?

lukecarbis:
:boom: Thanks @greguly

greguly:
You are welcome, it was my pleasure to help. :-)

PR #806, opened 2015-12-10
Only load scripts when on the correct admin pages

Stayallive:
Some plugins clash with the Select2 library and since the scripts are not needed on other pages then Stream admin pages we should not load them.

I am not sure if I overlooked something on this change, so it would be nice if a second pair of eyes checks this.

Would love to get some feedback in before I dig in and fix the test :)

### DIFF
--- a/classes/class-admin.php
+++ b/classes/class-admin.php
@@ -334,24 +334,24 @@ class Admin {
 	 * @return void
 	 */
 	public function admin_enqueue_scripts( $hook ) {
-		wp_register_script( 'select2', $this->plugin->locations['url'] . 'ui/lib/select2/select2.js', array( 'jquery' ), '3.5.2', true );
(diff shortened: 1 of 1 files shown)


lukecarbis:
@Stayallive registering the script isn't the same as loading it. I imagine it could cause problems if other plugins registered a script with the same name (likely a different version of the same thing).

Maybe the best thing to do here, instead of moving the `wp_register_script()` calls inside the page check, is to rename our script handlers to `wp-stream-select2` and `wp-stream-timeago`?

Even with the scripts registered inside the page check, we're not guaranteed zero conflicts. For example, if another plugin registered their same name script on every page, including our specific pages.

Stayallive:
Ah yes this would be a way better solution! Let me fix that :+1:

PR #811, opened 2015-12-18
Increase generic callback action arguments to 99

### LINKED ISSUE #809: Possible fix for Connector_GravityForms::callback_gform_post_note_added() error
Hey folks,

I'm getting this error while adding a note into GravityForms entry.
```
E_WARNING: Missing argument 6 for WP_Stream\Connector_GravityForms::callback_gform_post_note_added()
```

I get this message when I add a note using the following code which runs as a feed on gravity forms (not sure if that part is relevant). Stream does appear to log the action correctly though.
```
RGFormsModel::add_note( $lead_id, $user_id, $user_name, 'some message here', 'note' );
```

I narrowed it down to https://github.com/xwp/stream/blob/develop/classes/class-connector.php#L38 which adds the action to the 'gform_post_note_added' hook with 5 variables.
```
add_action( $action, array( $this, 'callback' ), 10, 5 );
```

However, the Gravity Forms do_action uses 6 variables in forms_model.php around lines 1309-1318
```
do_action( 'gform_post_note_added', $wpdb->insert_id, $lead_id, $user_id, $user_name, $note, $note_type );
```

Changing https://github.com/xwp/stream/blob/develop/classes/class-connector.php#L38 to 6 variables fixed the issue for me, however I do not know if that will cause any unintended consequences with other functions.

I hope that helps, or if there is another proper way to fix, let me know. I searched through the issues here and didn't see anything related.

EDIT: Forgot to mention WordPress 4.4, GravityForms 1.9.15, Stream 3.0.4

lukecarbis:
Resolves #809.

@elconejito - could you please test this and let me know how you go?

### DIFF
--- a/classes/class-connector.php
+++ b/classes/class-connector.php
@@ -35,7 +35,7 @@ abstract class Connector {
 	 */
 	public function register() {
 		foreach ( $this->actions as $action ) {
-			add_action( $action, array( $this, 'callback' ), 10, 5 );
+			add_action( $action, array( $this, 'callback' ), 10, 99 );
 		}
 
 		add_filter( 'wp_stream_action_links_' . $this->name, array( $this, 'action_links' ), 10, 2 );


elconejito:
Hey @lukecarbis I will give this a whirl when I'm back in the office on monday!

elconejito:
Tested this just now. It solves the original error and I don't see any ill effects. Logging of actions continues as expected. No other errors or warnings pop up.

lukecarbis:
:+1:

PR #812, opened 2015-12-21
Fix potentiel script name conflicts

Stayallive:
As discussed on PR #806 it would be better to rename the select2 and other third party libs to have a prefix to prevent the loading of scripts from another plugin.

### DIFF
--- a/classes/class-admin.php
+++ b/classes/class-admin.php
@@ -334,17 +334,17 @@ class Admin {
 	 * @return void
 	 */
 	public function admin_enqueue_scripts( $hook ) {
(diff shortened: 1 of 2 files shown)


Stayallive:
Travis failing on PHP 7.0 and WP 3.9 :)

lukecarbis:
Thanks for the PR @Stayallive. It's almost there. Can you please remove the `npm-debug.log` file?

Stayallive:
Dang, don't know how that sneaked in, excuse me :) It's gone now.

Issue #814, opened 2016-01-22
Role array without 0 index preventing log from being created


desrosj:
Working on a site and noticed debug.log instances where the following would occur:

```WordPress database error Column 'user_role' cannot be null for query INSERT INTO `wp_stream````

Looked around and tracked it this line in classes/class-log.php

```'user_role'  => (string) ! empty( $user->roles ) ? $user->roles[0] : '',```

After some debugging I found that the roles property of WP_User is not guaranteed to be an array starting with an index of 0. It could be any numerical value.

Attaching a change that prevents this from blocking the log item.

PR #815, opened 2016-01-22
Fixing bug where WP_User->roles array does not start with a 0 index

desrosj:
See https://github.com/xwp/stream/issues/814 for context.

### DIFF
--- a/classes/class-log.php
+++ b/classes/class-log.php
@@ -109,7 +109,7 @@ class Log {
 			'site_id'    => (int) is_multisite() ? get_current_site()->id : 1,
 			'blog_id'    => (int) apply_filters( 'wp_stream_blog_id_logged', get_current_blog_id() ),
 			'user_id'    => (int) $user_id,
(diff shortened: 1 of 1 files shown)


chacha:
@desrosj Attempting to reproduce the issue. Do you know under what circumstances you were not getting a 0 index array?

desrosj:
I believe I have it narrowed down to the user having a capability added to their account prior to their current role. The key in the roles property in WP_User appears to correspond with it's placement in the caps property. The site that I noticed this error on was assigning capabilities to users that were not in a role in order to allow access to content in a drip manner. 

To reproduce:
1. Create new user.
2. With code, or a plugin that allows you to add custom capabilities, add a capability that is not a role to the user (in our scenario it was `$user->add_cap( 'custom_cap' )`.
3. Change the user's role (can be with code or in the admin).

You should then notice the error then. If you print out or log the `WP_User` object, you will notice the order of items in `$user->caps` and the keys in `$user->roles`.

chacha:
Reproduced the bug and verified the code fix. Refreshing the commit in `pr/815/

PR #817, opened 2016-02-12
Exclude problematic builds from Travis

### LINKED ISSUE #816: Travis build failing
Currently the Travis build on all branches in failing. However, it doesn't appear like this has anything to do with Stream. Instead, it appears that WordPress 3.9 and PHP7 cause a large number of deprecated warnings due to the deprecated of the old constructor style:

    Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP; Translation_Entry has a deprecated constructor in /tmp/wordpress/src/wp-includes/pomo/entry.php on line 14

Because of this, it might make sense to remove this combination from the Travis test suite. I'll submit a pull request.

chacha:
Excludes WordPress 3.9 from being tested in conjunction with PHP7.

Errors are being caused because WordPress 3.9 still contains old constructors that are now deprecated. These deprecations are causing errors preventing the build from passing.

Closes #816

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,3 +32,10 @@ after_script:
     - $DEV_LIB_PATH/travis.after_script.sh
 
 sudo: false
+
+matrix:
+  exclude:
+    - php: 7.0
+      env: WP_VERSION=3.9 WP_MULTISITE=0
+    - php: 7.0
+      env: WP_VERSION=3.9 WP_MULTISITE=1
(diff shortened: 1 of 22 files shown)


chacha:
After re-running the latest `develop` branch test, it appears there are quite a few more problems than I originally saw in a past test. Going to start playing CodeSniffer Whack-a-Mole(tm) now.

chacha:
Awesome. We're all set. Looks like it was just some new Coding Standards along with incompatibilities between PHP and WordPress versions.

/cc @lukecarbis

lukecarbis:
Great work on this @chacha !

PR #823, opened 2016-02-29
CSV Export

### LINKED ISSUE #813: CSV Export
Feature request - would be great to export Stream data to a CSV file for analysis. For example, I want to know the top 10 most active user accounts by login to measure engagement. Currently I can only see the activity on a user-by-user basis so it would be great to quickly slice and dice the data in a more portable form.

chacha:
Closes #813 

Adds a CSV export download link. Allows user to download any view of Stream Records. Uses the currently selected filters.

Also adds `stream_records_per_page` filter.


chacha:
@lukecarbis The only thing I'm not quite happy with is the page item count seems to be pushed down. I've played around with it a bit, but haven't gotten very far. Suggestions?

classes/class-admin.php
```diff
@@ -687,6 +690,94 @@ public function render_list_table() {
 	<?php
 	}
 
+	public function maybe_render_csv_page() {
```
lukecarbis: I would suggest moving the `maybe_render_csv_page`, `render_csv_disable_paginate`, and `render_csv_expand_columns` into their own `Export` class, then instantiating it as a property of `Admin`.

classes/class-admin.php
```diff
@@ -687,6 +690,94 @@ public function render_list_table() {
...
+		$page = wp_stream_filter_input( INPUT_GET, 'page' );
+		$output_csv = wp_stream_filter_input( INPUT_GET, 'output_csv' );
+		if (  '1' !== $output_csv || 'wp_stream' !== $page ) {
```
lukecarbis: Rather than using the parameter `output_csv` with a value of `1`, can we make this more extendable by using: `output` with a value of `csv`? Might mean some refactoring.

classes/class-admin.php
```diff
@@ -687,6 +690,94 @@ public function render_list_table() {
...
+	 */
+	public function render_csv_expand_columns( $columns ) {
+		return array(
+			'date'      => $columns['date'],
+			'summary'   => $columns['summary'],
+			'user_id'   => $columns['user_id'],
+			'connector' => __( 'Connector', 'stream' ),
+			'context'   => $columns['context'],
+			'action'    => $columns['action'],
+			'blog_id'   => __( 'Blog ID', 'stream' ),
```
lukecarbis: Maybe only add this if `is_multisite() && is_plugin_active_for_network( $this->plugin->locations['plugin'] )`.

chacha: Good catch. I made a mental note to check when this is added regularly, but should have written it down.

classes/class-admin.php
chacha: Done. I agree that this spot is more suitable for it.

classes/class-admin.php
chacha: Done. Great suggestion.

classes/class-export.php
```diff
@@ -0,0 +1,118 @@
...
+					case 'ip' :
+						$row_out[] = $record->{$column_name};
+						break;
+				}
+			}
+			$csv_output[] = $row_out;
+		}
+
+		header( 'Content-type: text/csv' );
+		header( 'Content-Disposition: attachment; filename="output.csv"' );
```
lukecarbis: Let's brand it: stream.csv

lukecarbis:
Loving the progress on this. What I'd really like to see if Exporting modularised. I'm imagining an `Export` class, and an `Export_CSV` class.

The `Export` class would register all the export types (filterable), and include them in the Export options at the bottom of the list table.

The `Export_CSV` class would look after the pagination filter, columns, and outputting the CSV file.

Basically, what I'm getting at is that it should be very easy to create an `Export_JSON` class without touching or duplicating any existing code.

Make sense?

chacha:
Definitely. And we might as well add that while we’re at it.

I’m thinking about having `Export` build a data array that adds the base data (and could be filtered) that will then be passed to the specific exporter. That way anyone could add extra information without having to rely on individual exporter implementations.

chacha:
Something about the copy messed up all of the tabs/indents causing Travis to fail. I'm going to have to go through and individually fix it. Will do that a little later today.

lukecarbis:
Great thoughts. Let's do it.

classes/class-export.php
```diff
@@ -0,0 +1,136 @@
...
+
+		if ( 'wp_stream' === wp_stream_filter_input( INPUT_GET, 'page' ) ) {
+			add_action( 'admin_init', array( $this, 'render_download' ) );
+			add_filter( 'stream_exporters', array( $this, 'register_default_exporters' ) );
+		}
+
+	}
+
+	public function render_download() {
+
```
lukecarbis: Drop blank line at top of method.

classes/class-export-json.php
```diff
@@ -0,0 +1,16 @@
+<?php
+namespace WP_Stream;
+
+class Export_JSON {
+
+  public function output_file ( $data ) {
```
lukecarbis: Exporters should also have a `$name` or `$title` property (which is output on the admin screen). There should be an abstract `Exporter` class which each exporter extends (similar to our `Connector` class). The `Export` class should check if each registered exporter is a type of `Exporter`.

classes/class-list-table.php
```diff
@@ -545,6 +549,31 @@ public function get_filters() {
...
+		}
+		
+		foreach ( array_keys( $exporters ) as $export_type ) {
+			$args = array_merge( array( 'output' => $export_type ), $_GET );
+			$download = add_query_arg( $args, 'admin.php' );
+
+			$export_links[] = sprintf(
+				'<a href="%s">%s</a>',
+				esc_html( $download ),
+				esc_html( strtoupper ( $export_type ) )
```
lukecarbis: See comment above. This will change to something like `esc_html( $this->admin->export->exporters[ $export_type ]->name )`.

classes/class-admin.php
```diff
@@ -25,6 +25,11 @@ class Admin {
 	public $live_update;
 
 	/**
+	 * @var Export
+	 */
+	public $export;
```
lukecarbis: @chacha Question: Should Export belong to `Admin`? Or should it belong to `Plugin`? Maybe `List_Table`? Thoughts?

chacha: Right now the exporter specifically takes Stream Records from a List Table and Exports. So I would probably place it with `List_Table`.

lukecarbis: Is it possible to decouple `Exports` from `List_Table`?

chacha: It is possible. One option would be an adapter class between the two, or moving several of the methods we just migrated away from Admin back into admin.

lukecarbis: Since `Exports` relies on `List_Table` for filters, it doesn't really make sense. Let's keep it in `Admin`.

chacha:
Things that need to get done

- [ ] Fix Travis Build
- [ ] Add Document Blocks
- [ ] Add Unit Tests

chacha:
@lukecarbis Trying to add a test for the CSV/JSON Exporter's `output_file()` method, but a headers sent warning is preventing the test from being executed/success. Any thoughts/suggestions?

chacha:
@lukecarbis Other than the above test coverage conerns, I believe we're ready to roll.

lukecarbis:
@chacha You could use the `WP_STREAM_TESTS` global, to return something different, that works without a warning, but can still be checked for accuracy. See here for an example usage: https://github.com/xwp/stream/blob/39163bb4c74ae7b955e42f17ea8fe84006ed4b5d/classes/class-admin.php#L552

tests/tests/test-class-export.php
```diff
@@ -0,0 +1,83 @@
...
+		 $this->assertArrayHasKey( 'user_id', $columns );
+		 $this->assertArrayHasKey( 'connector', $columns );
+		 $this->assertArrayHasKey( 'context', $columns );
+		 $this->assertArrayHasKey( 'action', $columns );
+		 $this->assertArrayHasKey( 'ip', $columns );
+	 }
+
+	 /**
+	  * Test no output on normal page load
+		*/
```
lukecarbis: Spacing.

lukecarbis:
Let's make sure that we have one unit test for each method in `Export` and `Exporter`. Also, I would unit test our individual Exporters (CSV and JSON) separately.

chacha:
@lukecarbis Everything we've talked about is finished. I think we're ready to merge. Let me know if something is missing.

classes/class-exporter-json.php
```diff
@@ -0,0 +1,39 @@
...
+	 * @return void
+	 */
+	public function output_file( $data, $columns ) {
+
+		if ( ! defined( 'WP_STREAM_TESTS' ) || ( defined( 'WP_STREAM_TESTS' ) && ! WP_STREAM_TESTS ) ) {
+			header( 'Content-type: text/json' );
+			header( 'Content-Disposition: attachment; filename="stream.json"' );
+		}
+
+		if ( function_exists( 'wp_json_encode' ) ) {
```
lukecarbis: 😍

lukecarbis:
@chacha Could you please review my commits above, and give me your thoughts?

chacha:
@lukecarbis Reviewed all commits. I am all for bringing this inline with how connectors are registered. Consistency is important. 

 :+1:

classes/class-export.php
```diff
@@ -0,0 +1,242 @@
...
+	/**
+	 * Checks whether an exporter class is valid
+	 *
+	 * @param Exporter $exporter The class to check.
+	 * @return bool
+	 */
+	public function is_valid_exporter( $exporter ) {
+		if ( ! is_a( $exporter, 'WP_Stream\Exporter' ) ) {
+			trigger_error(
```
chacha: We should probably trigger this error in `register_exporters()`. You should be able to call `is_valid_exporter()` on anything and simply get true or false. Otherwise it might be disjarring, akin to raising an exception with you call `is_array( 'string' )`.

lukecarbis: :+1:

tests/tests/test-class-export.php
```diff
@@ -0,0 +1,177 @@
...
+	 * Test for present columns returning
+	 */
+	public function test_expand_columns() {
+		$test_data = array(
+		 'date'		=> '',
+		 'summary' => '',
+		 'user_id' => '',
+		 'context' => '',
+		 'action'	=> '',
+		 'ip'			=> '',
```
lukecarbis: Spacing needs to be fixed here.

PR #826, opened 2016-03-08
Remove border-bottom from inactive nav tabs

barryceelen:
A border is added to the inactive tabs on the settings page causing the tabs to display incorrectly.

pasted_image_08_03_16_20_04


lukecarbis:
Thank you!

PR #827, opened 2016-03-10
Logging changes of settings in Customizer

### LINKED ISSUE #308: Changes made in the Customizer are not being logged
After clicking **Save & Publish** in the Customizer no records are being logged by Stream for changes that were performed.

screen shot 2014-03-07 at 8 59 24 pm

We'll need to improve the following connectors so that these changes are always tracked:

- [x]  Widgets (#391)
- [ ] Menus
- [ ] Settings

/cc @westonruter

lukecarbis:
Resolves #308.

Thanks @marcin-lawrowski for finding this. I've tested it and it works well.

@westonruter Would you mind reviewing this one-liner? Any unintended side-effects?

### DIFF
--- a/connectors/class-connector-settings.php
+++ b/connectors/class-connector-settings.php
@@ -541,7 +541,7 @@ class Connector_Settings extends Connector {
 	}
 
 	/**
-	 * Trigger this connector from WP CLI, only for known Settings
+	 * Trigger this connector from WP CLI or the Customizer, only for known Settings
 	 *
 	 * @action update_option
 	 *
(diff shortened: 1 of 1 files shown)


westonruter:
Using the `customize_save` action is a good way to capture whether the setting updates are being made in the Customizer, yes: https://github.com/xwp/wordpress-develop/blob/deaabdcc28bfbdc78d90040ece81250563b84779/src/wp-includes/class-wp-customize-manager.php#L999-L1020

The PHPDoc will need to be updated from:

> Trigger this connector from WP CLI, only for known Settings

lukecarbis:
Good catch. Thanks @westonruter.

Issue #828, opened 2016-03-10
Add missing Settings strings


lukecarbis:
WordPress contains some settings which aren't being picked up properly by the Settings connector.

For example:
`nav_menu_locations `
`gmt_offset`
`default_link_category`

We need to identify every possible setting (possibly by `ack`ing WordPress core for `update_option` and `update_site_option`?) and add them to the Settings connector's `$labels` array.

In addition, we should identify all the Theme settings for all the default WordPress themes, from `twentyten` through to `twentysixteen` and add them to the Settings connector too. Alternatively, these could be added to a separate connector for themes, which essentially does the same thing.

PR #831, opened 2016-03-14
Theme settings for all the default WordPress themes

marcin-lawrowski:
Fixes https://github.com/xwp/stream/issues/828

### DIFF
--- a/connectors/class-connector-settings.php
+++ b/connectors/class-connector-settings.php
@@ -285,14 +285,6 @@ class Connector_Settings extends Connector {
 					'header_textcolor',
 				),
 			),
-
-			'twentyeleven_theme_options' => array(
-				'settings' => array(
-					'color_scheme',
-					'link_color',
-					'theme_layout',
-				),
-			),
 		);
 
 		if ( isset( $contexts[ $option_name ] ) ) {
@@ -406,12 +398,6 @@ class Connector_Settings extends Connector {
 				'secondary_text_color'    => esc_html__( 'Secondary Text Color', 'stream' ),
 				'link_color'              => esc_html__( 'Link Color', 'stream' ),
 				'page_background_color'   => esc_html__( 'Page Background Color', 'stream' ),
-			),
-
-			'twentyeleven_theme_options'  => array(
-				'color_scheme'            => esc_html__( 'Color Scheme', 'stream' ),
-				'link_color'              => esc_html__( 'Link Color', 'stream' ),
-				'theme_layout'            => esc_html__( 'Default Layout', 'stream' ),
 			),
 		);
 
@@ -577,8 +563,7 @@ class Connector_Settings extends Connector {
 	 * @param mixed $value
 	 */
 	public function callback_update_option( $option, $value, $old_value ) {
-		$is_theme_option = sprintf( '%s_theme_options', get_option( 'stylesheet' ) ) == $option;
-		if ( ( defined( '\WP_CLI' ) && \WP_CLI || did_action( 'customize_save' ) ) && ( $is_theme_option || array_key_exists( $option, $this->labels ) ) ) {
+		if ( ( defined( '\WP_CLI' ) && \WP_CLI || did_action( 'customize_save' ) ) && array_key_exists( $option, $this->labels ) ) {
 			$this->callback_updated_option( $option, $value, $old_value );
 		}
 	}
--- a/connectors/class-connector-settings.php
+++ b/connectors/class-connector-settings.php
@@ -322,6 +322,27 @@ class Connector_Settings extends Connector {
 	}
 
 	/**
+	 * Find out if the option should be ignored and not logged
+	 *
+	 * @param string $option_name
+	 *
+	 * @return bool Whether the option is ignored or not
+	 */
+	public function is_option_ignored( $option_name ) {
+		if ( 0 === strpos( $option_name, '_transient_' ) || 0 === strpos( $option_name, '_site_transient_' ) ) {
+			return true;
+		}
+
+		$ignored = array(
+			'image_default_link_type',
+			'medium_large_size_w',
+			'medium_large_size_h',
+		);
+
+		return in_array( $option_name, $ignored, true );
+	}
+
+	/**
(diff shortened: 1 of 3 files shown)


connectors/class-connector-settings.php
```diff
@@ -385,7 +395,24 @@ public function get_serialized_field_label( $option_name, $field_key ) {
...
+				'sidebar_textcolor'      => esc_html__( 'Header and Sidebar Text Color', 'stream' ),
+				// Custom Colors
+				'color_scheme'           => esc_html__( 'Color Scheme', 'stream' ),
+				'main_text_color'        => esc_html__( 'Main Text Color', 'stream' ),
+				'secondary_text_color'   => esc_html__( 'Secondary Text Color', 'stream' ),
+				'link_color'             => esc_html__( 'Link Color', 'stream' ),
+				'page_background_color'  => esc_html__( 'Page Background Color', 'stream' ),
 			),
+
+			'twentyeleven_theme_options' => array(
```
lukecarbis: Why do we include a second array of labels here? How are twentyeleven options different?

connectors/class-connector-settings.php
```diff
@@ -550,7 +577,8 @@ function( $rule ) use ( $submenu, $record ) {
 	 * @param mixed $value
 	 */
 	public function callback_update_option( $option, $value, $old_value ) {
-		if ( ( defined( '\WP_CLI' ) && \WP_CLI || did_action( 'customize_save' ) ) && array_key_exists( $option, $this->labels ) ) {
+		$is_theme_option = sprintf( '%s_theme_options', get_option( 'stylesheet' ) ) == $option;
+		if ( ( defined( '\WP_CLI' ) && \WP_CLI || did_action( 'customize_save' ) ) && ( $is_theme_option || array_key_exists( $option, $this->labels ) ) ) {
```
lukecarbis: This statement is looking pretty messy. Can we clean it up?

connectors/class-connector-settings.php
marcin-lawrowski: Twentyeleven theme has "Twenty Eleven Theme Options" page in Appearance menu and settings listed there are saved in "twentyeleven_theme_options" option

lukecarbis: What do you think of this - just saving a generic "Twenty Eleven Theme Options" setting - instead of going into detail?

I'm not sure it's worth it for the added code complication.

What do you think?

lukecarbis:
Got these two when updating Media settings (for the first time):
`medium_large_size_w`
`image_default_link_type`

What would they be referring to?

lukecarbis:
I also got _heaps_ of these ones, when updating the "Front Page displays" setting for the first time:

`"([0-9]{4})/([0-9]{1,2})/([0-9]{1,2})/page/?([0-9]{1,})/?$" setting was updated`

Can we exclude all settings that end with a `$`?

lukecarbis:
In the Stream settings (also part of the Settings adaptor), I'm getting duplicate entries for the Keep Records For Setting. It appears like this:

screen shot 2016-03-15 at 09 59 52

marcin-lawrowski:
Options "medium_large_size_w" and "image_default_link_type" are reset to empty string every time Media settings are saved because there is not input fields to edit them and they are present in $whitelist_options. I assume they are either internal or legacy. I wrote a code that ignores these options.

marcin-lawrowski:
As for the duplicated entries - when you select "Keep Records Indefinitely" then "Keep Records for" setting is cleared. When you clear  "Keep Records Indefinitely" checkbox then "Keep Records for"  setting is set to 30 days (default). So, when you update "Keep Records Indefinitely" setting then "Keep Records for" is updated as well.

connectors/class-connector-settings.php
```diff
@@ -322,6 +322,27 @@ public function is_key_ignored( $option_name, $key ) {
...
+	 */
+	public function is_option_ignored( $option_name ) {
+		if ( 0 === strpos( $option_name, '_transient_' ) || 0 === strpos( $option_name, '_site_transient_' ) ) {
+			return true;
+		}
+
+		$ignored = array(
+			'image_default_link_type',
+			'medium_large_size_w',
+			'medium_large_size_h',
```
lukecarbis: Is there such a thing as `small_medium_size_w` or other size variations?

connectors/class-connector-settings.php
```diff
@@ -333,6 +333,10 @@ public function is_option_ignored( $option_name ) {
 			return true;
 		}
 
+		if ( substr( $option_name, -1 ) === '$' ) {
```
lukecarbis: Yoda conditional, this should be.

lukecarbis:
@marcin-lawrowski Re: Duplicated entries:

Perhaps an easy and less confusing solution would be to simply change the label for "Keep Records for" to "Record TTL". The Stream Settings screen would remain the same. Thoughts?

chacha:
Just to pipe in...

> Perhaps an easy and less confusing solution would be to simply change the label for "Keep Records for" to "Record TTL".

I think that TTL is a technical phrase that I wouldn't expect many people to understand. 

>When you clear "Keep Records Indefinitely" checkbox then "Keep Records for" setting is set to 30 days (default). So, when you update "Keep Records Indefinitely" setting then "Keep Records for" is updated as well.

Two alternatives:

- Keep the value of "Keep Records for" and simply use JS to hide the option when indefinitely is selected. Internally just have the indefinite checkbox override.
- Don't store the indefinite value and instead, replace "Keep Records for" with a `0` to indicate an indefinite time period.

connectors/class-connector-settings.php
marcin-lawrowski: No, there is not. "medium_large_size_w" and "medium_large_size_w" are the only options in this case.

connectors/class-connector-settings.php
marcin-lawrowski: Done

marcin-lawrowski:
@chacha I also think that TTL phrase could be too difficult, so I decided to hide "Keep Records for" option if the checkbox is selected. Now there is only one Stream log when an user selects / deselects the checkbox.

Issue #829, opened 2016-03-13
Multisite: No way to exclude Super Admins?


clubdesign:
Hey Guys,

thanks for your great work, but we have a problem here, i can't exclude Super Admins from Tracking. We have 3 Super Admins on our site (60 Subsites) and i can't find them within the select on the exclude screen (The role is missing and these 3 Users are missing). All of the Super Admins Activity is tracked right now. I don`t know how you get the users right now, but Super Admins don't have to attached to a blog, so maybe the Users Query needs to be adjusted?

Is there any filter available to manually exclude those users? Can't find anything in the docs (and was too lazy yet to read through the code ;) ).

Greets Marcus

clubdesign:
Forget my Laziness, just went into code, as far as i can see, the only option currently is to hook into your `wp_stream_settings_options` filter and add the user_ids i need to exclude rules. That seems to work smoothly, but it should be made easier for this case ;)

```
add_filter("wp_stream_settings_options", function( $wp_stream_rules ) {

  $exclude_user_ids = array(1);

  if( array_key_exists("author_or_role", $wp_stream_rules["exclude_rules"] ) || !is_array( $wp_stream_rules["exclude_rules"]["author_or_role"] ) ) {
    $wp_stream_rules["exclude_rules"]["author_or_role"] = $exclude_user_ids;
  } else {
    $wp_stream_rules["exclude_rules"]["author_or_role"] = array_merge( $wp_stream_rules["exclude_rules"]["author_or_role"], $exclude_user_ids );
  }

  return $wp_stream_rules;

 }, 10, 1 );
```
By the way, thumbs up for your clean code, i wish other plugins work like that!

lukecarbis:
@marcin-lawrowski Could you look into this? It shouldn't be hard to do the following:

1. Check if the Stream is running on multisite
2. Check if the current user is a super admin, and if so
3. Include super admins in the dropdown list of users

Let's make sure this works on the Exclude screen as well as the User filter screen.

PR #835, opened 2016-03-15
Include super admins to the users list in Exclude settings

marcin-lawrowski:
Fixes https://github.com/xwp/stream/issues/829

It's a work in progress


classes/class-settings.php
```diff
@@ -114,12 +114,29 @@ public function get_users() {
 		if ( 0 === $users->get_total() ) {
 			wp_send_json_error( $response );
 		}
+		$users_array = $users->results;
+
+		if ( is_multisite() && is_super_admin() ) {
+			$super_admins = get_super_admins();
+			foreach ( $super_admins as $admin ) {
+				$user = get_user_by ( 'login', $admin );
+				$users_array[] = $user;
```
lukecarbis: Should we check if the user already exists in the array first?

classes/class-settings.php
```diff
@@ -114,12 +114,29 @@ public function get_users() {
...
 		$response->message = '';
 		$response->users   = array();
+		$users_added_to_response = array();
+
+		foreach ( $users_array as $key => $user ) {
+			// exclude duplications:
+			if ( array_key_exists( $user->ID, $users_added_to_response ) ) {
+				continue;
+			} else {
+				$users_added_to_response[$user->ID] = true;
```
lukecarbis: Spacing around the variable inside the square brackets.

classes/class-settings.php
marcin-lawrowski: I choose to check for duplicates further in the foreach loop.

classes/class-settings.php
marcin-lawrowski: Fixed

lukecarbis:
@chacha I've made a few changes in f4a600d. Could you please review, and if you're happy, merge?

classes/class-list-table.php
```diff
@@ -464,10 +464,14 @@ function( $user_id ) {
 			);
 
 			if ( is_multisite() && is_super_admin() ) {
-				$super_admins = get_super_admins();
-				foreach ( $super_admins as $admin ) {
-					$users[] = get_user_by( 'login', $admin );
```
lukecarbis: This array is an array of `Author`s, not `WP_User`s. Hence the change.

classes/class-list-table.php
```diff
@@ -464,10 +464,14 @@ function( $user_id ) {
...
-				}
+				$super_admins = array_map(
+					function( $login ) {
+						$user = get_user_by( 'login', $login );
+						return new Author( $user->ID );
+					},
+					get_super_admins()
+				);
+				$users = array_unique( array_merge( $users, $super_admins ) );
```
lukecarbis: Array unique so we don't have duplicates if a Super Admin is also an Administrator for the site.

chacha:
Tested. Good Job @marcin-lawrowski! :+1:

PR #844, opened 2016-04-19
Stream Alerts Alpha

chacha:
With a good amount of the work laid out, it is time to start honing in on the final prototype and what needs to get done. Here's a basic list:

Core Functionality
- [x] Allow notifiers to display/save form information
- [ ] Add unit tests to test all core API functionality
- [x] Allow users to active and deactivate alerts (`post_status`?)

User Interface
- [x] Polish 'Edit Alert' page CSS/HTML.
- [x] Make sure all quick edit information is relevant.
- [x] Tweak Publish Box to only display relevant entries

Alert Types
- [x] Finish Email Message Implementation
- [x] Add Highlight Notifier
- [x] Add Emailer Notifier
- [x] Add IFTTT Notifier

Bugs
- [x] Alerts list cannot be seen on Network Admin
- [x] Status missing from Alerts list table
- [x] Default states of Trigger dropdowns is wrong (should be the same as Exclude Rules)
- [x] Clearing the Trigger dropdowns gives a javascript error and doesn't work
- [x] Currently active option should default to enabled
- [x] Remove hover state actions in the "matched records" section
- [x] Highlights can't be cleared
- [x] "Records matching these features" does not update automatically

### DIFF
--- a/classes/class-alert.php
+++ b/classes/class-alert.php
@@ -237,4 +237,131 @@ class Alert {
 	public function send_alert( $record_id, $recordarr ) {
 		$this->get_alert_type_obj()->alert( $record_id, $recordarr, $this );
 	}
+
+	public static function get_record_triggered_alerts( $record, $alert_slug, $single_alert_id ) {
+		if ( ! is_string( $alert_slug ) || ! is_numeric( $single_alert_id ) ) {
+			return false;
+		}
+
+		if ( is_array( $record ) ) {
+			$record = (object) $record;
+		}
+		if ( empty( $record->ID ) ) {
+			return false;
+		}
+		$record = new Record( $record );
+		$alerts_triggered = $record->get_meta( Alerts::ALERTS_TRIGGERED_META_KEY, true );
+
+		if ( empty( $alerts_triggered ) || ! is_array( $alerts_triggered ) ) {
+			$alerts_triggered = array( $alert_slug => array( $single_alert_id ) );
+		} else if ( ! array_key_exists( $alert_slug, $alerts_triggered ) || ! is_array( $alerts_triggered[ $alert_slug ] ) ) {
+			$alerts_triggered[ $alert_slug ] = array( $single_alert_id );
+		} else if ( ! in_array( $single_alert_id, $alerts_triggered[ $alert_slug ] ) ) {
+			$alerts_triggered[ $alert_slug ][] = $single_alert_id;
+		}
+		return $record->update_meta( Alerts::ALERTS_TRIGGERED_META_KEY, $alerts_triggered );
+	}
+
+	/**
+	 * Record Alerts triggered by a Record.
+	 * 
+	 * This should be used any time an alert is triggered.
+	 *
+	 * Stores the post ID of an Alert triggered by a record.
+	 *
+	 * As an example, this exists so that its value(s) can then be used
+	 * to fetch meta from that Alert in later operations.
+	 *
+	 * This also creates the Alert triggered meta if not found.
+	 *
+	 * @see Alert_Type_Highlight::alert() for an example.
+	 *
+	 * @param object     $record The Record.
+	 * @param string     $alert_slug The Alert Type slug.
+	 * @param int|string $single_alert_id The single Alert ID.
+	 * @return bool If the meta was updated successfully.
+	 */
+	public static function update_record_triggered_alerts( $record, $alert_slug, $single_alert_id ) {
+		if ( ! is_string( $alert_slug ) || ! is_numeric( $single_alert_id ) ) {
+			return false;
+		}
+
+		if ( is_array( $record ) ) {
+			$record = (object) $record;
+		}
+		if ( empty( $record->ID ) ) {
+			return false;
+		}
+		$record = new Record( $record );
+		$alerts_triggered = $record->get_meta( Alerts::ALERTS_TRIGGERED_META_KEY, true );
+
+		if ( empty( $alerts_triggered ) || ! is_array( $alerts_triggered ) ) {
+			$alerts_triggered = array( $alert_slug => array( $single_alert_id ) );
+		} else if ( ! array_key_exists( $alert_slug, $alerts_triggered ) || ! is_array( $alerts_triggered[ $alert_slug ] ) ) {
+			$alerts_triggered[ $alert_slug ] = array( $single_alert_id );
+		} else if ( ! in_array( $single_alert_id, $alerts_triggered[ $alert_slug ] ) ) {
+			$alerts_triggered[ $alert_slug ][] = $single_alert_id;
+		}
+		return $record->update_meta( Alerts::ALERTS_TRIGGERED_META_KEY, $alerts_triggered );
+	}
+
+	/**
+	 * Get a meta value from the Alert that a Record has triggered.
+	 *
+	 * If a Record has triggered an Alert (post), this fetches a specific
+	 * Alert meta (i.e., "post meta") value from that Alert.
+	 *
+	 * First, it gets the array of Alerts that the Record has triggered.
+	 * Then, using the requested Alert Type, it grabs the first item (post ID) in
+	 * that Type.
+	 *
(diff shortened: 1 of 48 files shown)


classes/class-alert.php
```diff
@@ -0,0 +1,144 @@
+<?php
+namespace WP_Stream;
+
+class Alert {
+	public $ID;
```
lukecarbis: Please add docblocks.

classes/class-alert.php
```diff
@@ -0,0 +1,144 @@
...
+
+	public $filter_author;
+	public $filter_action;
+	public $filter_context;
+	public $alert_type;
+	public $alert_meta;
+
+	public $notifier;
+
+	public function __construct( $item ) {
```
lukecarbis: Docblocks for all the functions, too. :)

classes/class-alerts.php
```diff
@@ -0,0 +1,472 @@
...
+	* @var Plugin
+	*/
+	public $plugin;
+
+	/**
+	* Post meta prefix
+	*
+	* @var string
+	*/
+	public $meta_prefix = 'wp_stream';
```
lukecarbis: This seems unused.

lukecarbis:
I'd prefer not to use the term "Notifier" in relation to Alerts. Instead, can we rename the class, etc. to `Alert_Type`.

classes/class-alert.php
```diff
@@ -0,0 +1,144 @@
+<?php
+namespace WP_Stream;
+
+class Alert {
+	public $ID;
+	public $date;
+	public $author;
+
+	public $filter_author;
```
lukecarbis: Instead of the term `filter`, let's use `trigger`.

classes/class-preview-list-table.php
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	/**
+	 * Display the table
+	 *
+	 * @since 3.1.0
+	 * @access public
+	 */
+	public function display() {
+		$singular = $this->_args['singular'];
```
lukecarbis: Is this unused?

classes/class-form-generator.php
```diff
@@ -0,0 +1,91 @@
+<?php
+namespace WP_Stream;
+
+class Form_Generator {
```
lukecarbis: Let's make sure this gets used by settings.

lukecarbis:
In the same way an Alert has an Alert Type (currently called a notifier), I feel that the Alert Trigger should be abstracted as well.

Currently, the trigger is set to Author, Context, and Action. Rather than limiting it to these three, it should be possible to write a "Trigger" extension to also trigger by IP Address.

As such, the folder hierarchy and naming conventions should be changed to:
```
- alerts
    - class-alert-type-highlight.php
    - class-alert-type-email.php
    - class-alert-trigger-author.php
    - class-alert-trigger-connector.php
    - class-alert-trigger-context.php
    - class-alert-trigger-action.php
```

classes/class-alert.php
chacha: Will switch then over when abstracting out all of the trigger types.

chacha:
We could eventually make triggers work in two modes. The first one is pretty standard and uses the 3-4 main default triggers. We could also eventually allow an advanced mode that treats triggers more like the rules in the Exclude section of settings.

chacha:
@lukecarbis I'd love to get your opinion on the title logic here. I updated the current language to be in this format that made sense to me:

    {Alert Type} when {User} {action} an item in {Context}
    Email when Tyler attached an item in Posts
    Highlight when Luke activated an item in Plugins

These are the problems I see with this:

- It would be useful if actions had past and future tense labels. Displaying a record should reference an action is the past ('Tyler updated') but for rules, future tense makes more sense ('Tyler updates')
- 'item' would be much better if it were replaced or removed. In reality, I'd want to see something like 'Email when Tyler activated a plugin', or 'Highlight when Luke attached images in Posts'. This would probably require a lot of testing and finding edge cases, but is doable. 
- We are still using context in a hierarchy. I know last time we talked, the idea of the action being performed 'inside of' a context wasn't ideal. Trying to figure out how best to express that.

The ideal system would be to get to a language processor that could create lines like:

- Email when Tyler activates a theme.
- Highlight when Luke installs a plugin.
- Email when Tyler updates a page. 
- Create Menu Alert when Tyler updates settings.
- Highlight when Luke creates a new post.

lukecarbis:
Looking good!

1. Switch the Actions and Contexts menus
2. The Stream menu should be highlighted when adding a new Alert
3. When saving an Alert for the first time, it is saved as "Published" instead of "Enabled"
4. Add "Bulk Actions" > "Move to Trash" (or "Restore") to Alerts list table
5. The list table inside "Records matching these triggers" isn't responsive
6. There's an extra pipe (`|`) after the status filters on the Alerts list table

lukecarbis:
On the topic of title logic:

I've put a lot of thought into this, and come to the conclusion that natural language is not the best way forward. Here's why:
1. It's hard to parse visually
2. It borders on impossible to translate

Instead, I put together this mockup. What do you think?
screen shot 2016-05-17 at 13 36 09

We would just need to add a `$color` property to each trigger.

chacha:
@lukecarbis On title: I like the concept. I'm going to give it a try on a branch and see how it works out.

chacha:
@lukecarbis If you get a chance, check out `title-revamp`, and let me know what you think with that implementation.

lukecarbis:
Added some styles to `title-revamp`. Looking good.

I think we should swap the Type and Trigger columns.

chacha:
@lukecarbis I tried switching the two columns, and thought it created a left column that was very heavy. I tried adding some more information about the alert in the Type column. Take a look at `alerts-descriptive-types`.

lukecarbis:
Need to go through and update all the docblocks.

chacha:
Currently the only reason this isn't passing is because of a bug that is already fixed in another branch.

lukecarbis:
@chacha Had a quick look, found a few things that aren't yet completed, and some bugs. Added them to the description of this PR.

johnregan3:
@lukecarbis @jacobschweitzer  This would be a good point to review my work on how Highlights work, including that Record meta field that tracks what Alerts have been triggered.

johnregan3:
At this point, the Network Admin menu link to the Alerts should be fixed.  I've written it so that if we're on the Network Admin menu, use the link to the first Site Admin, where the Alerts that apply to the Network Records are truly located.

johnregan3:
Hmmm.  I'm looking at this again, and I see an issue -- the Network interface could be complicated.

So, what's currently happening is that on the Network Stream, only Records from the first network site are being highlighted.

To reproduce:

1. Create a Highlight alert on the first network site (the main admin that is not the Network admin: `wp-admin/edit.php?post_type=wp_stream_alerts`). Set it to Yellow.

2. Go to another network site and create a highlight alert and set it to Red.

3. Trigger the Alert on both sites several times.

4. Navigate to the Network Stream Records list.

You'll only see yellow highlights, and they'll only be on records from the first network site.

screen shot 2016-08-04 at 9 18 22 pm

lukecarbis:
This is looking great! Only two things I found worth bringing up:

1. I feel that the IFTTT connector might need some instructions / extra setup information. @jwold, could you please take a look and make a suggestion here?

2. The email notification email wording needs some work. The subject is "Whatever".

> You've received a Stream Alert.
> 
> Action: unapproved
> Connector: comments
> Triggered By: admin
> 
> The alert is in reference to the following post:
> 
> ID: 1
> Title: Hello world!
> Last Updated: 2016-08-02 04:42:56

@jwold Again, could you make a suggestion on improving tone, language, style, information, etc.? Also, I think it should include a call to action, too.

jwold:
@lukecarbis I'll take a look at these two and suggest a change.

lukecarbis:
@jacobschweitzer Could you please review c9a45dd?

jacobschweitzer:
Looks good @lukecarbis !

PR #845, opened 2016-04-23
Remove get_meta and update_meta from Record

chacha:
Doesn't appear to work as intended. A look at the database shows that it isn't actually saved.

### DIFF
--- a/classes/class-record.php
+++ b/classes/class-record.php
@@ -59,31 +59,6 @@ class Record {
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


lukecarbis:
These functions are used in a bunch of connectors - won't this break them?

Might be better to fix the functions, rather than removing them.

chacha:
Immediately it would appear that they are probably working. For some reason, they weren't functioning throughout my testing. I will look into this.

PR #851, opened 2016-05-26
Fix order status log for WooCommerce >= 2.2

jamesgol:
Per https://woocommerce.wordpress.com/2014/08/07/wc-2-2-order-statuses-plugin-compatibility/ WooCommerce switched from using a taxonomy for shop_order_status to using post_status.

### DIFF
--- a/connectors/class-connector-woocommerce.php
+++ b/connectors/class-connector-woocommerce.php
@@ -54,6 +54,8 @@ class Connector_Woocommerce extends Connector {
 
 	private $settings = array();
 
+	private $plugin_version = null;
+
 	public function register() {
 		parent::register();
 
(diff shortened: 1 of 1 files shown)


jamesgol:
While trying to understand how everything works in the plugin I decided to look at some existing connectors and noticed this wasn't working on my recent version of WooCommerce.

jamesgol:
I'm not convinced that strtolower is desired on the names or that the status should be passed as the log action.  Might be nicer to pass 'updated' as the action.

lukecarbis:
Thanks @jamesgol - see https://github.com/xwp/stream/commit/4b46c5593e43f0e9e70b59881d5f5546119045c4 and https://github.com/xwp/stream/commit/9560e3c429fbdc87b738f5d31ee82b9c99ac20cf

Issue #852, opened 2016-05-31
Exclude Rules interface not working after Select2 update


lukecarbis:
```
The select2('destroy') method was called on an element that is not using Select2.
```

chacha:
In reference, this is the [documentation](https://select2.github.io/announcements-4.0.html#migrating-from-select2-35) for the biggest problems caused by moving to Select2 4.0+. 

By no longer supporting hidden input tags, it means that the entire page has to be rewritten for regular `<select>` boxes. This means a complete rehaul of the Javascript hitting the excludes page.

Given the size of the exclude page's JS, I'm going to separate it into a JS file specific for excludes and hopefully do some reworking so it is simpler. Right now there is a lot going on in different areas of the file.

chacha:
The switch creates an issue when we are using `<optgroup>`s instead of a hierarchy through the `<input type="hidden" />` methodology. The most realistic way to keep this functionality is to have connectors be a part of the regular options and style it differently to create the same effect.

lukecarbis:
What's the issue with `<optgroup>`s?

chacha:
In past versions you could select a group, like I could select the entire
Posts context, because it was do through a JSON object instead of through
select options. This was never possible with the select HTML format. With
the upgrade, they have not included any augmentations to allow you to
select optgroups, making that feature unobtainable. I believe there was an
official response speaking to this on an issue in the Select2 github.

lukecarbis:
Link?

I'm not sure we need to select a whole group. Just the connector, or the context.

chacha:
In previous versions, the connector _was_ the whole group, as it is a hierarchy.

The Select 2 issue is select2/select2#520.

lkraav:
I'm trying to add the `N/A` user to the excluded list, but can't because of this bug I think.

chacha:
@lkraav I see that problem as well. I am looking into it now.

PR #854, opened 2016-06-03
pass ajax parameter to filter_select

evgenykireev:


lukecarbis:
Thanks @evgenykireev . 

@chacha could you please look into this as a possible fix for #852?

chacha:
We can use this as a base for loading users through a Select2 Box. This has obvious benefits (quick search, etc).

PR #855, opened 2016-06-08
Select2 Upgrade/Exclude List 

chacha:
Moves towards using `<select>` boxes in the entire exclude interface.

@lukecarbis Please give feedback on any bugs you see.

See #852


lukecarbis:
Some issues I've found:

- [ ] Context Menu
    - [ ] Top level items should be bold
    - [ ] Filtering by top level items doesn't show the item pre-selected in the menu
    - [ ] Indents, instead of dashes, for secondary level items
    - [ ] Items with the same name (e.g. Installer > Plugins and Editor > Plugins) cause problems
- [ ] Need to strip params from the `_wp_http_referer` before the form is submitted, otherwise you end up with really long URLs really quick
- [ ] Exclude Rules default state should be "Any Author", "Any Context", etc.
- [ ] Exclude Rules don't actually have any effect (Tested with Context: Comments and Action: Approved) 
- [ ] IP Address field
    - [ ] Too high (doesn't match other fields)
    - [ ] Highlight state doesn't match
    - [ ] Doesn't allow inputting custom IPs

classes/class-form-generator.php
```diff
@@ -0,0 +1,200 @@
...
+	 */
+	public function render_all() {
+		$output = '';
+		foreach ( $this->fields as $data ) {
+			$output .= $this->render_field( $data['type'], $data['args'] );
+		}
+		return $output;
+	}
+
+	public function render_as_table() {
```
lukecarbis: Rename to `render_fields_table()`

classes/class-form-generator.php
```diff
@@ -0,0 +1,200 @@
...
+			'args' => $args,
+		);
+	}
+
+	/**
+	 * Renders all fields currently registered.
+	 *
+	 * @return string
+	 */
+	public function render_all() {
```
lukecarbis: Rename to `render_fields()`

classes/class-form-generator.php
lukecarbis: Docblock

classes/class-form-generator.php
```diff
@@ -0,0 +1,200 @@
...
+	 * @var array
+	 */
+	public $fields = array();
+
+	/**
+	 * Class constructor
+	 *
+	 * @return void
+	 */
+	public function __construct() {
```
lukecarbis: Not required, remove.

classes/class-form-generator.php
```diff
@@ -0,0 +1,200 @@
...
+		return $output;
+	}
+
+	/**
+	 * Renders a single field.
+	 *
+	 * @param string $field_type The type of field being rendered.
+	 * @param array  $original_args The options for the field type.
+	 */
+	public function render_field( $field_type, $original_args ) {
```
lukecarbis: Change `$original_args` to `$args`

classes/class-form-generator.php
```diff
@@ -0,0 +1,200 @@
...
+				$output = apply_filters( 'wp_stream_form_render_field', $output, $field_type, $original_args );
+				break;
+		}
+
+		$output .= ! empty( $args['description'] ) ? wp_kses_post( sprintf( '<p class="description">%s</p>', $args['description'] ) ) : null;
+
+		return $output;
+	}
+
+	public function prepare_data_string( $data ) {
```
lukecarbis: Docblock

lukecarbis: Rename to `prepare_select2_data_string`? Not 100% clear what this method does.

classes/class-settings.php
```diff
@@ -809,21 +810,22 @@ public function render_field( $field ) {
...
-						esc_attr( wp_create_nonce( 'stream_get_users' ) )
-					);
+					$author_or_role_input = $form->render_field( 'select2', array(
+						'name'        => esc_attr( sprintf( '%1$s[%2$s_%3$s][%4$s][]' , $option_key, $section, $name, 'author_or_role' ) ),
+						'options'     => $author_or_role_values,
+						'classes'     => 'author_or_role',
+						'data'        => array(
+							'group'         => 'connector',
+							'placeholder'   => esc_html__( 'Any Author or Role', 'stream' ),
+							'nonce'         => esc_attr( wp_create_nonce( 'stream_get_users' ) ),
```
lukecarbis: Is this creating a nonce per input? We only want one nonce per form.

classes/class-form-generator.php
chacha: This method prepares the HTML data attributes key/value pairs. For instance, if you have an array with `array( 'my-key' => 'my-value' )` it will turn it into an HTML data attribute: `data-my-key="my-value"`. 

This is used to output all of the data attributes on the form elements.

chacha:
@lukecarbis There were a couple incongruencies that I was unable to figure out.

> Need to strip params from the _wp_http_referer before the form is submitted, otherwise you end up with really long URLs really quick

After monitoring the `_wp_http_referer` parameter, it never is any longer than `/wp-admin/admin.php?page=wp_stream_settings`.

> Top level items should be bold
> Indents, instead of dashes, for secondary level items

Looking at the exclude page, I do not see the problems you are talking about. See below:

screen1

I'll be looking to see if there is a way to reproduce the problems you're finding.

lukecarbis:
In regards to the `_wp_http_referer` param, you can replicate by filtering by action, then changing the action filter and pressing the Filter button again. Repeat 4 - 5 times and you will get a URL too long error.

lukecarbis:
In regards to the Top level items should be bold, Indents, instead of dashes… This occurs in the filters on the Stream screen.

classes/class-settings.php
lukecarbis: @chacha

lukecarbis:
@chacha Please review all the code, and test. This should now fix everything.

chacha:
@lukecarbis Looks great.

Issue #836, opened 2016-03-17
Custom Roles not tracking


lukecarbis:
To replicate:

1. Create a custom user role (any plugin should do)
2. Activate the User Switching plugin (https://wordpress.org/plugins/user-switching/)
3. Create and then switch to a user with custom role
4. Update post draft
5. Switch back to admin user
6. Update post draft again

You will only see entries for the admin user, not the user with a custom role. It seems as though custom roles aren't having any activity records logged.

Reported here: https://wordpress.org/support/topic/custom-roles-not-tracking?replies=4

marcin-lawrowski:
I tried to reproduce this issue but no effect. I am using WP 4.4.2 and the newest Stream. I created a few custom roles and users for those roles. All activities of those users were properly tracked in Stream either with User Switching plugin or without it.

marcin-lawrowski:
@lukecarbis Have you tried to replicate it?

lukecarbis:
I just did, and I can't.

<img width="1066" alt="screen shot 2016-03-22 at 11 19 51" src="https://cloud.githubusercontent.com/assets/1097667/13939344/134dfb10-f020-11e5-8586-8672042f9eac.png">

lukecarbis:
@joelworsham Could there be something else going on here? Do you happen to have an object cache?

joelworsham:
Yes, we use WP Engine, so there is some caching going on.

lukecarbis:
Try clearing your object cache. I think the problem is that you're creating a new role, but the object cache isn't cleared. When Stream tries to get the user's role, and unknown role is returned, which results in a silent fail. Not ideal, but can you test by:

1. Create a new role
2. Try logging activities in that role (and see that they don't log)
3. Clearing your cache
4. Try logging activities in that role again

joelworsham:
Hmmm I tried that and it still does not work. I've recorded a screencast just so you can see it with your own :eyes: 

https://www.dropbox.com/s/3oquayjpm4ghpgo/Stream%20Custom%20Roles%20Example%20-%203-22-16.mov?dl=0

lukecarbis:
Thanks for helping us try to get to the bottom of this, Joel. I can't replicate the same thing on my end at all. Maybe could you post the code you're using to create the user role in a gist? I'll try out the same role with the same capabilities, etc. and see if I can replicate it that way.

joelworsham:
Sure thing.

https://gist.github.com/joelworsham/f64990299e061aea9896

[...]

PR #862, opened 2016-07-22
Bugfix: #836 Custom Roles not tracking

rheinardkorf:
Addresses https://github.com/xwp/stream/issues/836 where longer custom user rolls are not getting entries into the database.

The column size of 20 characters does not pass string validation for user rolls with longer lengths and causes a WP_Error and in turn a `wp_stream_record_insert_error`.

### DIFF
--- a/classes/class-install.php
+++ b/classes/class-install.php
@@ -399,7 +399,7 @@ class Install {
 			blog_id bigint(20) unsigned NOT NULL DEFAULT '1',
 			object_id bigint(20) unsigned NULL,
 			user_id bigint(20) unsigned NOT NULL DEFAULT '0',
-			user_role varchar(20) NOT NULL DEFAULT '',
+			user_role varchar(50) NOT NULL DEFAULT '',
 			summary longtext NOT NULL,
 			created datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
 			connector varchar(100) NOT NULL,
--- a/classes/class-install.php
+++ b/classes/class-install.php
@@ -337,6 +337,7 @@ class Install {
 		$db_update_versions = array(
 			'3.0.0' /* @version 3.0.0 Drop the stream_context table, changes to stream table */,
 			'3.0.2' /* @version 3.0.2 Fix uppercase values in stream table, connector column */,
+			'3.0.8' /* @version 3.0.8 Increase size of user role IDs, user_roll column */,
 		);
 
 		/**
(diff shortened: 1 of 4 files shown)


rheinardkorf:
I added a DB update for version 3.0.8, but if it doesn't make it to this version the update function needs to be renamed.

includes/db-updates.php
```diff
@@ -1,5 +1,22 @@
...
+ * Version 3.0.8
+ *
+ * Force update for older versions to call \dbdelta in install() method to fix column widths.
+ *
+ * @param string $db_version
+ * @param string $current_version
+ *
+ * @return string
+ */
+function wp_stream_update_308( $db_version, $current_version ) {
```
lukecarbis: Since this is a small change, and won't take long to execute, we should make it an auto update (so the user doesn't need to press a button to update the tables). If you agree, the function name just needs to change to `wp_stream_auto_update_308`.

rheinardkorf: Agreed. I'll update the method and update the PR.

lukecarbis: Not sure if you'll need to update this or not: https://github.com/xwp/stream/blob/develop/classes/class-install.php#L337-L340

classes/class-install.php
```diff
@@ -337,6 +337,7 @@ public function db_update_versions() {
 		$db_update_versions = array(
 			'3.0.0' /* @version 3.0.0 Drop the stream_context table, changes to stream table */,
 			'3.0.2' /* @version 3.0.2 Fix uppercase values in stream table, connector column */,
+			'3.0.8' /* @version 3.0.8 Increase size of user role IDs, user_roll column */,
```
lukecarbis: I was wrong about this - it isn't needed for auto update versions.

lukecarbis:
This is great. Just tested, works well, except for the comment I left just now.

@rheinardkorf Can you please also add some unit tests to the record logging to make sure really long entries in all fields always work? Once there's a unit test, I'll merge.

classes/class-install.php
rheinardkorf: Sure thing. I'll remove this.

rheinardkorf:
@lukecarbis sure I'll do some unit tests.

classes/class-install.php
rheinardkorf: I think it is required. The `update` method runs through the versions returned by `db_update_versions` and then handles the update one at a time.  Unless I am missing a method somewhere that will call the update function.

lukecarbis: Just tested, and you are right. There is a bug, however, which makes the "Update Database" dialog show, even when there's only auto updates to run.

PR #869, opened 2016-08-16
Update actions dropdowns for exclusions based on context.

jacobschweitzer:
Allows for Ajax to update the actions dropdown when adding a new exclude rule. The JavaScript could be added to admin.js or global.js and PHP functions could be made general so that it could be use on any Stream settings page. As of right now this is just a first pass at getting the functionality working.

### DIFF
--- a/classes/class-connectors.php
+++ b/classes/class-connectors.php
@@ -120,25 +120,25 @@ class Connectors {
 
 		foreach ( $this->connectors as $connector ) {
 			if ( ! method_exists( $connector, 'get_label' ) ) {
-				$this->plugin->admin->notice( sprintf( __( "%s class wasn't loaded because it doesn't implement the get_label method.", 'stream' ), $connector->name, 'Connector' ), true );
+				$this->plugin->admin->notice( sprintf( __( '%s class wasn\'t loaded because it doesn\'t implement the get_label method.', 'stream' ), $connector->name, 'Connector' ), true );
 				continue;
 			}
 			if ( ! method_exists( $connector, 'register' ) ) {
-				$this->plugin->admin->notice( sprintf( __( "%s class wasn't loaded because it doesn't implement the register method.", 'stream' ), $connector->name, 'Connector' ), true );
+				$this->plugin->admin->notice( sprintf( __( '%s class wasn\'t loaded because it doesn\'t implement the register method.', 'stream' ), $connector->name, 'Connector' ), true );
 				continue;
 			}
 			if ( ! method_exists( $connector, 'get_context_labels' ) ) {
-				$this->plugin->admin->notice( sprintf( __( "%s class wasn't loaded because it doesn't implement the get_context_labels method.", 'stream' ), $connector->name, 'Connector' ), true );
+				$this->plugin->admin->notice( sprintf( __( '%s class wasn\'t loaded because it doesn\'t implement the get_context_labels method.', 'stream' ), $connector->name, 'Connector' ), true );
 				continue;
 			}
 			if ( ! method_exists( $connector, 'get_action_labels' ) ) {
-				$this->plugin->admin->notice( sprintf( __( "%s class wasn't loaded because it doesn't implement the get_action_labels method.", 'stream' ), $connector->name, 'Connector' ), true );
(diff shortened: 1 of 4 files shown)


classes/class-settings.php
```diff
@@ -1061,4 +1064,21 @@ public function get_settings_translations( $labels ) {
...
+		$actions = $GLOBALS['wp_stream']->connectors->term_labels['stream_action'];
+		if ( ! empty( $connector ) ) {
+			$class_name = '\WP_Stream\Connector_'.ucfirst( $connector );
+			if ( class_exists( $class_name ) ) {
+				$connector_class = new $class_name();
+				$actions         = $connector_class->get_action_labels();
+			}
+		}
+		ksort( $actions );
+		die( wp_json_encode( $actions ) );
```
lukecarbis: Can we use `wp_send_json_success()` instead here, and drop the `die()`?

The js would obviously need to be updated to support this, too.

classes/class-settings.php
```diff
@@ -1068,17 +1068,24 @@ public function get_settings_translations( $labels ) {
...
+	function get_actions() {
+		$connector_name = wp_stream_filter_input( INPUT_POST, 'connector' );
+
+		if ( empty( $connector_name ) ) {
+			wp_send_json_error();
+		}
+
+		$actions = wp_stream_get_instance()->connectors->term_labels['stream_action'];
+
+		$connector_class = '\WP_Stream\Connector_' . ucfirst( $connector_name );
```
lukecarbis: This won't work for multiname connectors, such as `Connector_User_Switching`, and it won't work if the connector's `name` property doesn't match it's class name.

classes/class-connectors.php
```diff
@@ -93,7 +93,7 @@ public function load_connectors() {
 				continue;
 			}
 			if ( $class->is_dependency_satisfied() ) {
-				$classes[] = $class;
+				$classes[ $class->name ] = $class;
```
lukecarbis: This will mean that, unfortunately, third-party extension initialising classes without the class name as the key will not get the benefit of auto-filtering the Actions menu.

lukecarbis: I still think it's the best way forward, though.

lukecarbis:
@jacobschweitzer Looking good. I've made a few changes. Could you please review, and leave your thoughts? If you're happy, we can merge, and start thinking about the Actions menu in the Stream list table filters.

jacobschweitzer:
@lukecarbis  This looks great, thank you for these updates/comments! I was thinking about making the connectors more accessible with an associative array in the global but this works better, I didn't want to add bloat to the global or modify the structure then have to deal with issues. 

I'd taken a look at the actions menu in the Stream list table yesterday but it is a bit different than the exclusions or alerts. This new code will certainly help though, I'll take a fresh look at it again today.

ui/js/admin.js
```diff
@@ -498,7 +499,56 @@ jQuery( function( $ ) {
...
+
+		var placeholder = $('<option/>', {value: '', text: ''});
+		action_select.append( placeholder );
+		var data = {
+			'action'    : 'get_actions',
+			'connector' : connector
+		};
+		$.post( window.ajaxurl, data, function( response ) {
+			var success = response.success,
+				actions = response.data;
```
lukecarbis: Use spaces, not tabs, when aligning variables like this.
```
var success = x,
    actions = y;
```

ui/js/admin.js
```diff
@@ -498,7 +499,56 @@ jQuery( function( $ ) {
 			}
 		});
 	});
+	var updateFilterActions = function( element ) {
+		var connector = 'Any';
+		var connector_element = $( 'select[name="context"]' );
+		var select2_data = connector_element.select2('data');
+		if ( undefined !== select2_data[0].element['attributes']['data-group'] ) {
```
lukecarbis: Can this be alternately expressed as:
`select2_data[0].data( 'group' )`?

Or maybe:
`select2_data[0].attr( 'data-group' )`?

ui/js/admin.js
```diff
@@ -498,7 +499,56 @@ jQuery( function( $ ) {
 			}
 		});
 	});
+	var updateFilterActions = function( element ) {
```
lukecarbis: I don't think we need to define the `element` param here.

```
var updateFilterActions = function() {
```

jacobschweitzer:
This has been merged into #844

PR #872, opened 2016-08-26
Create add new alert inline on alerts listing page.

jacobschweitzer:
This is a rough (working) draft for adding new alerts on the alerts listing page instead of using a new post page. A lot of the existing functions relied on the fact that the post object was already present but in this case we need to create it so some functions had to be changed and some new ones added to show the form fields. 

Here is a quick demo:
http://i.imgur.com/BPzvtBX.gifv

This is still a bit sloppy: the form fields could be brought in via AJAX (right now hidden form fields are brought in from the footer), the new alert could be added to the list without refreshing, and some code could be consolidated / improved. I just wanted to get it out there for people to have a look or help out.


lukecarbis:
This work has been merged into https://github.com/xwp/stream/pull/873

PR #877, opened 2016-09-12
Issue 876 - Keep For setting # of days is hidden

jacobschweitzer:
This pull request brings back the setting for keeping Stream records for a certain number of days. I've tested changing the network setting which works to purge records after applying the changes in this pull request. Testing should be done on the cron and site settings before merging this back in.

### DIFF
--- a/ui/js/settings.js
+++ b/ui/js/settings.js
@@ -1,8 +1,13 @@
 /* globals confirm, wp_stream, ajaxurl, wp_stream_regenerate_alt_rows */
 jQuery( function( $ ) {
-
-	var keepRecordsIndefinitely = $( '#wp_stream\\[general_keep_records_indefinitely\\]' ),
-	    keepRecordsFor          = $( '#wp_stream_general_records_ttl' ),
+	var network_affix;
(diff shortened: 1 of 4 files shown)


ui/js/settings.js
```diff
@@ -1,8 +1,13 @@
...
-
-	var keepRecordsIndefinitely = $( '#wp_stream\\[general_keep_records_indefinitely\\]' ),
-	    keepRecordsFor          = $( '#wp_stream_general_records_ttl' ),
+	var network_affix;
+	if ( 'wp_stream_network' === $( 'input[name="option_page"]' ).val() ) {
+		network_affix = '_network_affix';
+	} else {
+		network_affix = '';
+	}
+	var keepRecordsIndefinitely = $( '#wp_stream' + network_affix + '\\[general_keep_records_indefinitely\\]' ),
```
lukecarbis: Could you explain the forward slashes used here?

jacobschweitzer: I'll check on that, it is not a change I made, I just added the network affix part.

lukecarbis: Right! I see that. Would be good to understand why it's there though.

PR #865, opened 2016-08-02
Preparing Stream for Alternate Database Drivers

faishal:
- Enable alternate drivers in `WP_Stream\Plugin::__construct()`, Replaced the hard-coded `WP_Stream\DB` reference with a value that can be filtered, defaulting to the WordPress Database.
- Created `WP_Stream\DB_Driver` interface, which help to implement Driver classes
- Created `WP_Stream\DB_Driver_WPDB` class and implemented `WP_Stream\DB_Driver` interface, Moved all WordPress database related operation and query into driver class.
- Updated `WP_Stream\DB` class to use driver passed in constructor. Class contains wrapper function to communicate with driver, Moved all query related function into database driver file.
- Updated `WP_Stream\List_Table` to use `WP_Stream\DB` wrapper method instead of `WP_Stream\Query` class methods.
- Some minnor changes in other files to adopt new DB class methods.
- Added related test cases.

PR #881, opened 2016-09-15
Preparing Stream for Alternate Database Drivers

dkotter:
This is a continuation of PR #865. I'm going to be taking this over and seeing it though, but I don't have access to the original repo that was used for that PR, so I've pushed those same changes to the 10up fork, made the requested changes, and am continuing that PR here.

@lukecarbis I believe all the code formatting comments you made have been taken care of, other then the one mentioned here: https://github.com/xwp/stream/pull/865#discussion-diff-75428971. The issue on that one is the `prepare` method is used to prepare actual values, so for instance to prepare a `meta_value` you want to search for. It will automatically add single quotes around values.

So in this case, if we pass the query like that through prepare, it breaks the entire query, as it adds single quotes around all the statements which makes it not valid SQL. Now there might be a better way to handle this but the few methods I tried didn't work, so I've left as-is for now.

As for the 3 main things you mentioned:
- [ ] We should have a discussion about the Query class, and whether it should be renamed, or merged with DB_Driver_WPDB **(So not sure on this one, as I didn't write the original code here. If you have any strong opinions on this being renamed or merged, we can talk that through. I think it's fine where it's at but I'm also not opposed to making changes.)**
- [x] There are some TODO methods in DB_Driver_WPDB which still need implementation (purge_storage, setup_storage) **(These methods have been implemented. Note that the behavior shouldn't be any different as before but by implementing these, it allows alternate drivers to more easily run their own install/uninstall methods)** 
- [x] Some unit tests are failing **(Believe this is fixed. I ran unit tests locally and didn't get any failures)**

But let me know if you see other things that need fixed before this can be merged. Thanks!


classes/class-plugin.php
```diff
@@ -73,14 +73,24 @@ public function __construct() {
...
 		if ( ! $this->db ) {
+			$error = esc_html__( 'Stream: Could not load chosen DB driver.', 'stream' );
+		} elseif ( ! $driver instanceof DB_Driver ) {
+			$error = esc_html__( 'Stream: DB driver must implement DB Driver interface.', 'stream' );
+		}
+
+		if ( $error ) {
 			wp_die(
-				esc_html__( 'Stream: Could not load chosen DB driver.', 'stream' ),
+				esc_html( $error ),
```
lukecarbis: Can't use a variable as the first arg of `esc_html__()`.

Instead:
```
		if ( ! $this->db ) {
			$error = __( 'Stream: Could not load chosen DB driver.', 'stream' );
		} elseif ( ! $driver instanceof DB_Driver ) {
			$error = __( 'Stream: DB driver must implement DB Driver interface.', 'stream' );
		}
```

and then just:

```
				esc_html( $error ),
```

lukecarbis:
@dkotter Could you please give me access to your forked version, so I can work on this. Couple of little things that I can help with - and also I'm happy to merge that conflicting class if you like.

dkotter:
@lukecarbis I've added you to the fork. I can look into the merge conflict tomorrow if you don't get to it before then. Thanks!

dkotter:
@lukecarbis Any update here? Happy to help out if there's anything I can do.

lukecarbis:
@dkotter I'm away this week, but I've schedule a review of this for when I'm back on Monday.

PR #890, opened 2016-11-02
Prevent some connectors from loading on the site frontend

### LINKED ISSUE #878: Performance issue: 70+ Stream queries per page load when not logged in
I was profiling [goshen.edu](https://www.goshen.edu)'s server-side performance today using the [Query Monitor plugin](https://wordpress.org/plugins/query-monitor/) and noticed that Stream consistently makes SELECT queries, regardless of whether or not I'm logged in:

image

This is puzzling, as my understanding was that Stream only tracked the interactions of logged-in users. Sorry if I'm missing something obvious, but I thought I should pass this on.

lukecarbis:
Resolves #878.

@pranksinatra - can you please test this an confirm?

### DIFF
--- a/classes/class-connector.php
+++ b/classes/class-connector.php
@@ -31,6 +31,20 @@ abstract class Connector {
 	public $prev_stream = null;
 
 	/**
+	 * Register connector in the WP Admin
+	 *
+	 * @var bool
+	 */
+	public $register_admin = true;
+
+	/**
+	 * Register connector in the WP Frontend
+	 *
+	 * @var bool
+	 */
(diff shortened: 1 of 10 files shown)


pranksinatra:
@lukecarbis, thanks for looking into this and sorry for providing info on the specific queries sooner!

I tested the bugfix and unfortunately it didn't result in a significant decrease in queries (only removed 2 of the ~80). I did some digging and will reply on #878 in more detail.

lukecarbis:
@pranksinatra Thanks for the digging. I just pushed an update. Could you try again?

PR #892, opened 2016-11-02
Fix exclude rules

### LINKED ISSUE #876: Keep records option missing in Multisite
From the WordPress.org forums: [Keep records option missing in Multisite](https://wordpress.org/support/topic/keep-records-option-missing-in-multisite/)

> Hi
> I know there was a bug that removed the “keep records for xxx days” option which was resolved Version 3.0.7. But it’s still missing in multisite, unless its in a totally different place in multisite?

### LINKED ISSUE #888: Can't set up Exclude rules with 3.1.1 on Multisite
Adding new `Exclude` rules is basically impossible with 3.1.1 (possible 3.1, too) on Multisite as the `Context` and `Action` fields are not populated any longer (the `Author or Role` field is populated only with users, the `IP Address` field seems to work normally).

lukecarbis:
Resolves #888 
Resolves #876

@mundschenk-at Can you please test this out for me, make sure it works?

### DIFF
--- a/alerts/class-alert-trigger-author.php
+++ b/alerts/class-alert-trigger-author.php
@@ -107,8 +107,9 @@ class Alert_Trigger_Author extends Alert_Trigger {
 
 		foreach ( $users as $user ) {
 			$all_records[] = array(
-				'id'   => $user->id,
-				'text' => $user->get_display_name(),
+				'id'    => $user->id,
+				'value' => $user->id,
+				'text'  => $user->get_display_name(),
 			);
 		}
 		return $all_records;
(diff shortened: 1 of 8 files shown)


mundschenk-at:
Looks good. (The only thing not working quite right is the `Context` dropdown. There are some duplicate entries like "Settings". I think this is because the dropdown does not distinguish between "Connector" and "Context", so it's not actually related to issue #888, but a separate UI thing.)

lukecarbis:
@mundschenk-at They should be separated stylistically - that is, connector is bold, context isn't. Screenshot attached. Is this the case for you? Make sure you clear cache.

screen shot 2016-11-07 at 13 11 55

mundschenk-at:
No, it looks like a flat list on Firefox 45.3.0, even after clearing the cache. I'll check on Chrome/Safari tonight.

mundschenk-at:
Sorry for not getting back earlier. At least on OSX, both Chrome and Safari looked like this:
bildschirmfoto 2016-11-30 um 08 44 56

I haven't yet tried the fix in d5727a94e78887873a5500d6d5574babebbcf435.

lukecarbis:
@mundschenk-at Could you please pull the latest commits? I think this should sort things out.

mundschenk-at:
Yes, the lists now look like they should.

PR #893, opened 2016-11-04
Fix "trim() expects parameter 1 to be string, array given"

frozzare:
Sometimes `wp_stream_filter_input` returns a array than a string and you can't use `trim` with array. This pull request will solve that problem.

skarmklipp 2016-11-04 14 19 08


lukecarbis:
Thanks so much for this fix @frozzare, but I'm closing it in favour of #892.

PR #911, opened 2017-03-15
Up priority of driver set

tlovett1:
This makes it so ElasticPress can change the driver.


tlovett1:
@lukecarbis any idea when this will make it to a version?

PR #915, opened 2017-04-01
Add filter for settings access

davelozier:
The wp_stream_disable_admin_access filter is no longer present.  The new stream_allow_settings_access filter is named ambiguously so that using a white list or black list of usernames makes sense either way.  (or what ever else a developer wants to use)

### DIFF
--- a/classes/class-admin.php
+++ b/classes/class-admin.php
@@ -114,6 +114,13 @@ class Admin {
 	public $notices = array();
 
 	/**
+	 * Limit settings access via filter.
+	 *
+	 * @var bool
+	 */
+	private $allow_settings_access = true;
+
+	/**
 	 * Class constructor.
 	 *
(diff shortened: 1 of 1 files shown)


stephenharris:
Should maybe the permission be filtered instead?

stephenharris:
Here's a workaround for removing access to the settings page:

```
add_action( 'admin_menu', function(){
    //Re-add the page - with 'do_not_allow' capabilities - to 'update' the permission requirements
    //remove_submenu_page only removes the admin menu link
    add_submenu_page( 'wp_stream', '', '', 'do_not_allow', 'wp_stream_settings' );
    remove_submenu_page( 'wp_stream', 'wp_stream_settings' );
}, 999 );
```

Note that the `wp-admin/options.php` might still be available, but in this instance, the settings are serialised in the database, and so not editable there.

Issue #935, opened 2017-08-21
Undefined Index ID in customizer on post delete with customize-posts plugin


sayedtaqui:
When trashing a post in customizer with customize-posts plugin, stream causes this error
```
Notice: Undefined index: ID in /srv/www/wp.dev/docroot/wp-content/plugins/stream/classes/class-alerts-list.php on line 341
```

Its because customize-posts plugin is invoking `wp_insert_post_data` filter again and overriding `$postarr` to `$unsanitized_post_data`. 
https://github.com/xwp/wp-customize-posts/blob/develop/php/class-wp-customize-post-setting.php#L474

So when `Alerts_List::save_alert_inline_edit` hooked to `wp_insert_post_data` is called it does not get `$postarr['ID']` hence causing the above error.

https://github.com/xwp/stream/blob/develop/classes/class-alerts-list.php#L341

PR #936, opened 2017-08-21
Fix undefined index ID issue when trashing post with customize-posts …

sayedtaqui:
Fixes https://github.com/xwp/stream/issues/935


jeffpaul:
@lukecarbis note that this fix helps resolve downstream issues for Customize Posts as well as a Team Prometheus client.

Issue #954, opened 2017-11-21
'is_network_activated' should be true if stream is must-use


cfoellmann:
I would like to run stream as a must-use plugin. When I drop it in `mu-plugins` and require the stream.php it gets loaded on every load but it will not be detected as `is_network_activated() = true`.
```php
	/**
	 * Returns true if Stream is network activated, otherwise false
	 *
	 * @return bool
	 */
	public function is_network_activated() {
		if ( ! function_exists( 'is_plugin_active_for_network' ) ) {
			require_once( ABSPATH . '/wp-admin/includes/plugin.php' );
		}

		return is_plugin_active_for_network( $this->plugin->locations['plugin'] );
	}
```

2 solutions:

1. add a filter to set the return manually in is_network_activated() #955 
    - add stream to mu-plugins
    - `require_one 'stream/stream.php';` in your mu-plugins loader file or subfolder loader
    - `add_filter('wp_stream_is_network_activated', '__return_true');`
2. add a check like `is_mustuse()` #956

PR #956, opened 2017-11-21
add is_mustuse() detection

cfoellmann:
Alternative (potentially better) solution to #954


DavidCramer:
@cfoellmann I believe this is the best way since it's essentially automatic and won't cause frustration if docs are missed.

Issue #949, opened 2017-10-25
count_users() is called for every user role on the settings screen.


johnbillion:
On the Stream Settings screen, `count_users()` is unnecessarily called once for every role on the site, instead of just once. Calling `count_users()` is expensive, and gets slower the more users and roles there are on a site.

Ref: https://github.com/xwp/stream/blob/06316e9cc295429ae4ab7c550cddce898919e650/classes/class-settings.php#L821-L826

This call to `count_users()` should be moved outside of the `foreach()` loop so it's only called once.

DavidCramer:
adressed via #964

PR #957, opened 2017-12-03
Move `count_users` out of roles foreach

frozzare:
This pull request will move `count_users` out of roles foreach since it's expensive to call for every role. See #949 for more information.


DavidCramer:
@frozzare Thanks for this. Could you please change the PR against `develop` rather than `master`? 

Many thanks!

PR #963, opened 2018-03-06
Support proxy mode by checking the `X-Forwarded-For` header first

fjarrett:
@lukecarbis When in reverse proxy mode, the `REMOTE_ADDR` value will always show the IP of the proxy server (ATS, Varnish, HAProxy, etc).  Most proxies will pass along the real client IP via the `X-Forwarded-For` header.  If it exists, Stream should prefer that as the real IP and only use `REMOTE_ADDR` as a fallback.

👋 It's been a while 😃

### DIFF
--- a/classes/class-log.php
+++ b/classes/class-log.php
@@ -112,18 +112,18 @@ class Log {
 		$ip_address = $ip_address ? $ip_address : wp_stream_filter_input( INPUT_SERVER, 'REMOTE_ADDR', FILTER_VALIDATE_IP );
 
 		$recordarr = array(
-			'object_id'  => (int) $object_id,
-			'site_id'    => (int) is_multisite() ? get_current_site()->id : 1,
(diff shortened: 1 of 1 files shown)


classes/class-log.php
```diff
@@ -107,6 +107,10 @@ function( $var ) {
 			$role = '';
 		}
 
+		// Support proxy mode by checking the `X-Forwarded-For` header first
+		$ip_address = wp_stream_filter_input( INPUT_SERVER, 'HTTP_X_FORWARDED_FOR', FILTER_VALIDATE_IP );
+		$ip_address = $ip_address ? $ip_address : wp_stream_filter_input( INPUT_SERVER, 'REMOTE_ADDR', FILTER_VALIDATE_IP );
```
DavidCramer: @fjarrett This is cool, thanks. The `REMOTE_ADDR` is also requested on line 154/158 So will need to be applied there. 
Perhaps it might be better to have method to fetch the IP, or even a simpleIP param set in the constructor to prevent code duplication.

PR #964, opened 2018-04-03
Move `count_users` out of roles foreach

frozzare:
This pull request will move `count_users` out of roles foreach since it's expensive to call for every role. See #949 for more information.

This time against develop instead of master.

### DIFF
--- a/classes/class-settings.php
+++ b/classes/class-settings.php
@@ -771,8 +771,8 @@ class Settings {
 
 				break;
 			case 'rule_list':
-				$users = count_users();
-				$form = new Form_Generator();
+				$users  = count_users();
+				$form   = new Form_Generator();
 				$output = '<p class="description">' . esc_html( $description ) . '</p>';
 
 				$actions_top    = sprintf( '<input type="button" class="button" id="%1$s_new_rule" value="&#43; %2$s" />', esc_attr( $section . '_' . $name ),  esc_html__( 'Add New Rule', 'stream' ) );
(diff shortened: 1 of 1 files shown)


classes/class-settings.php
```diff
@@ -817,13 +817,14 @@ public function render_field( $field ) {
 					// Author or Role dropdown menu
 					$author_or_role_values   = array();
 					$author_or_role_selected = array();
+					$users = count_users();
```
DavidCramer: @frozzare You can move this even higher up, outside the `foreach` of the `exclude_row`. Perhaps place it at the beginning of the `rule_list` case.

frozzare:
Done!

classes/class-settings.php
```diff
@@ -771,6 +771,7 @@ public function render_field( $field ) {
 
 				break;
 			case 'rule_list':
+				$users = count_users();
 				$form = new Form_Generator();
 				$output = '<p class="description">' . esc_html( $description ) . '</p>';
```
DavidCramer: Cool. Could you please align the `=` so the build can padd ;)

classes/class-settings.php
```diff
@@ -823,7 +824,7 @@ public function render_field( $field ) {
 							'value' => $role_id,
 							'text' => $role,
 						);
-						$users = count_users();
+
```
DavidCramer: Also, clear up this line ;)

PR #972, opened 2018-04-06
Make reset nonce unique

### LINKED ISSUE #971: Use a unique nonce for the Reset Stream Database functionality
This is a follow-up to #945.

The nonce that's used for the Reset Stream Database functionality is the same as the nonce used for the Uninstall link on the Plugins listing screen.

This means that even if the Reset Stream Database functionality settings screen field is removed, it's still possible for a rogue user to trigger the database reset by manually calling `admin-ajax.php?action=wp_stream_reset&wp_stream_nonce=<nonce>`, where the nonce value is copied from the Uninstall link on the Plugins screen.

The Reset Stream Database functionality and the Uninstall functionality shouldn't used the same nonce.

DavidCramer:
This makes the nonce for the reset setting field unique to prevent manually sending a reset based on existing nonce.

Fixes #971

### DIFF
--- a/classes/class-admin.php
+++ b/classes/class-admin.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace WP_Stream;
 
 use DateTime;
@@ -8,6 +9,7 @@ use \WP_CLI;
 use \WP_Roles;
 
 class Admin {
+
 	/**
 	 * Hold Plugin class
 	 *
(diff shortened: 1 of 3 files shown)


DavidCramer:
props to @johnbillion for picking this up.

@lukecarbis feel free to review when ready.

PR #958, opened 2017-12-22
Fix busted IN, NOT IN query system [1]

lkraav:
Trying to use any `__in` or `__not_in` results in a database error:

```
WordPress database error You have an error in your SQL syntax; check the
manual that corresponds to your MariaDB server version for the right syntax to
use near ''context' IN ('')
```

It looks like the algorithm here has been busted for years. `array_shift()`
mutates the underlying `$values` array, losing data user has passed into the
query parameters.

This required rewriting some of the algorithm and throwing some stuff out.

I modelled the fix mostly by grepping for `prepare.*\\\ IN\\\ ` for examples
and picked [1]

EXAMPLE QUERY (custom connector `posts-viewed`)
```
$stream = wp_stream_get_instance();

 $records = $stream->db->get_records( [
     'action' => 'viewed',
     'author' => get_current_user_id(),
     'connector' => 'posts-viewed',
     'context__in' => [ 'course', 'lesson' ],
 ] );
```

[1]: https://github.com/WordPress/WordPress/blob/4.9.1/wp-includes/class-wp-comment-query.php#L794

PR #977, opened 2018-04-23
Fix busted IN, NOT IN query system [1]

lkraav:
(retrying #958)

Trying to use any `__in` or `__not_in` results in a database error:

```
WordPress database error You have an error in your SQL syntax; check the
manual that corresponds to your MariaDB server version for the right syntax to
use near ''context' IN ('')
```

It looks like the algorithm here has been busted for years. `array_shift()`
mutates the underlying `$values` array, losing data user has passed into the
query parameters.

This required rewriting some of the algorithm and throwing some stuff out.

I modelled the fix mostly by grepping for `prepare.*\\\ IN\\\ ` for examples
and picked [1]

EXAMPLE QUERY (custom connector `posts-viewed`)
```
$stream = wp_stream_get_instance();

 $records = $stream->db->get_records( [
     'action' => 'viewed',
     'author' => get_current_user_id(),
     'connector' => 'posts-viewed',
     'context__in' => [ 'course', 'lesson' ],
 ] );
```

[1]: https://github.com/WordPress/WordPress/blob/4.9.1/wp-includes/class-wp-comment-query.php#L794

### DIFF
--- a/classes/class-query.php
+++ b/classes/class-query.php
@@ -118,12 +118,10 @@ class Query {
 
 				$field = str_replace( array( 'record_', '__in' ), '', $key );
 				$field = empty( $field ) ? 'ID' : $field;
-				$type  = is_numeric( array_shift( $value ) ) ? '%d' : '%s';
+				$type  = is_numeric( $value[ 0 ] ) ? '%d' : '%s';
 
-				if ( ! empty( $value ) ) {
-					$format = '(' . join( ',', array_fill( 0, count( $value ), $type ) ) . ')';
-					$where .= $wpdb->prepare( " AND $wpdb->stream.%s IN {$format}", $field, $value ); // @codingStandardsIgnoreLine prepare okay
-				}
+				$format = '(' . join( ',', array_fill( 0, count( $value ), $type ) ) . ')';
+				$where .= $wpdb->prepare( " AND $wpdb->stream.{$field} IN {$format}", $value ); // @codingStandardsIgnoreLine prepare okay
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


classes/class-query.php
```diff
@@ -118,12 +118,10 @@ public function query( $args ) {
 
 				$field = str_replace( array( 'record_', '__in' ), '', $key );
 				$field = empty( $field ) ? 'ID' : $field;
-				$type  = is_numeric( array_shift( $value ) ) ? '%d' : '%s';
+				$type  = is_numeric( $value[ 0 ] ) ? '%d' : '%s';
```
lukecarbis: This won't work if the first value in the array isn't `0`. Could we use this as a non-mutating solution, instead?
`array_shift( array_values( $value ) );`

classes/class-query.php
```diff
@@ -146,12 +144,10 @@ public function query( $args ) {
 
 				$field = str_replace( array( 'record_', '__not_in' ), '', $key );
 				$field = empty( $field ) ? 'ID' : $field;
-				$type  = is_numeric( array_shift( $value ) ) ? '%d' : '%s';
+				$type  = is_numeric( $value[ 0 ] ) ? '%d' : '%s';
```
lukecarbis: Same here – can we use the non-mutating solution above?

classes/class-query.php
lkraav: But in what case wouldn't the first item's array index be exactly 0? Can there be string array index values, or?

I've yet to see any PHP warnings or anything in the 1 year this has been running in production. Wouldn't there be?

lkraav:
#WCEU notes w/ @kasparsd 

* sanitize prepare statement `$field` name vs table field names, because prepare will single-quote field name and generate invalid SQL
* `$value` needs to handle arrays for `IN` statement. `array_values()` suggestion by @lukecarbis seems to work well for this purpose.

classes/class-query.php
```diff
@@ -118,12 +118,10 @@ public function query( $args ) {
 
 				$field = str_replace( array( 'record_', '__in' ), '', $key );
 				$field = empty( $field ) ? 'ID' : $field;
-				$type  = is_numeric( array_shift( $value ) ) ? '%d' : '%s';
+				$type  = is_numeric( $value[ 0 ] ) ? '%d' : '%s';
 
-				if ( ! empty( $value ) ) {
-					$format = '(' . join( ',', array_fill( 0, count( $value ), $type ) ) . ')';
-					$where .= $wpdb->prepare( " AND $wpdb->stream.%s IN {$format}", $field, $value ); // @codingStandardsIgnoreLine prepare okay
```
kasparsd: `$wpdb->stream.%s` always becomes `$wpdb->stream.'FIELDNAME'` which never works. We need to specify the field name picking it from a list of known "property" field names and use it directly as `$wpdb->stream.{$field}`.

PR #987, opened 2018-07-04
Sanitise `$_SERVER` variable instead of escaping it.

fklein-lu:


lkraav:
What's the rationale?

fklein-lu:
The rationale is the code is doing it wrong. `esc_attr()`is for escaping. Here we are processing incoming data, which needs to be sanitised.

PR #993, opened 2018-12-12
Add support for mercator domain mapping plugin

spacedmonkey:
[Mercator](https://github.com/humanmade/Mercator) is a popular domain mapping plugin rewritten by Human made. This plugin is used by a number of VIP GO / high scale projects. This drop-in, should be logged by stream as well.

### DIFF
--- a/classes/class-connectors.php
+++ b/classes/class-connectors.php
@@ -76,6 +76,7 @@ class Connectors {
 			'edd',
 			'gravityforms',
 			'jetpack',
+			'mercator',
 			'user-switching',
 			'woocommerce',
 			'wordpress-seo',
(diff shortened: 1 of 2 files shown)


spacedmonkey:
Any chance this gets merged / review 
cc @lukecarbis @shadyvb

shadyvb:
LGTM!

Also check https://github.com/xwp/stream/blob/develop/connectors/class-connector-edd.php#L350 for how Stream dealt with very similar actions like the ones here, essentially where `action` is the only difference in the log message.

spacedmonkey:
I looked at a custom connector, but as mercator WordPress drop-in, there were issues around order of loading. I am using a custom connector right now, so I am using in production right now. Just thought I would share the code with XWP. 👍

PR #1003, opened 2019-06-17
Allow is_network_activated to be filtered

joehoyle:
Currently there's no way to filter the network activated detection (bar filtering the option plugins activated value). This means unless the `is_mustuse` is triggered, you can't get the network wide behaviour when loading Stream in other ways.

I refactored this into a method that is consistenty used on the `Plugin` class, and also added a filter. Forcing network behaviour is then just a case of doing `add_filter( 'wp_stream_is_network_activated', '__return_true' );`


kasparsd:
Thanks so much @joehoyle! Is this ready for code review?

joehoyle:
Yup, good stuff!

rheinardkorf:
Very elegant! Thanks @joehoyle :)

PR #1011, opened 2019-06-24
Fix IN and NOT IN queries

kasparsd:
Fixes #977.


lkraav:
Sup @kasparsd anything I can do to help here? Where did it get stuck?

kasparsd:
@lkraav I started working on this before we had the dev environment bundled with the plugin which can run unit tests, so the remaining job is to add the unit tests and do some smoke testing with some actual data.

kidunot89:
@kasparsd All ready for your review.

PR #1016, opened 2019-07-11
Tooling update and WPCS cleanup

kasparsd:
- [x] Switch to `eslint` to match WP core.
- [x] Enable the PHP coding standard checks.
- [x] Add a local dev environment.
- [x] Enable phpunit test reporting with Coveralls.


.codeclimate.yml
```diff
@@ -1,64 +0,0 @@
----
```
kasparsd: Removing since we no longer have access to this.

.eslintrc.json
```diff
@@ -0,0 +1,13 @@
+{
```
kasparsd: Switch to eslint.

Issue #959, opened 2018-01-10
CRON check is incorrect


khromov:
Before showing the "WP Cron Tracking" option in the admin, Stream checks if `DISABLE_WP_CRON` is not true.

I don't think this is a correct condition to check. `DISABLE_WP_CRON` doesn't actually disable CRON, it just stops it from running during normal requests. CRON still runs when manually loading `/wp-cron.php`. 

The check should be removed, as WP CRON can never be truly disabled.

DavidCramer:
@khromov Having thought this through, I believe that how it is now is in fact the correct way to do it.
When you define `DISABLE_WP_CRON`, you are telling WordPress to not use [`spawn_cron()`](https://core.trac.wordpress.org/browser/tags/4.9/src/wp-includes/cron.php#L371)

The `wp-cron.php` file you mention is the file used when you want to control the cron manually via your own method, as indicated in the files header in [`wp-cron.php`](https://core.trac.wordpress.org/browser/tags/4.9.5/src/wp-cron.php) :

> Defining DISABLE_WP_CRON as true and calling this file directly are
mutually exclusive and the latter does not rely on the former to work.

kasparsd:
I think the option to track cron events shouldn't be tied to the state of `DISABLE_WP_CRON` since that applies to the front-end CRON feature only and not cron jobs in general.

kasparsd:
Version 3.4.1 has been released, thanks for reporting the issue @khromov!

PR #1028, opened 2019-07-22
1013 Add cron check filter

### LINKED ISSUE #1013: Add filter for cron check
We use WP Stream, but have replaced native WP Cron with [HM Cavalcade](https://github.com/humanmade/Cavalcade). Part of installing cavalcade is setting `DISABLE_WP_CRON` to false (and allow cavalcade to run cron in parallel instead.

However, the [cron check](https://github.com/xwp/stream/blob/dd2f5809236d4849967a8ad860548a45b981ba4f/includes/functions.php#L123) in stream doesn't allow us to override this setting.

While #959 makes sense for general usage, I'd like to suggest creating a new filter to override this setting and allow compatibility with Cavalcade or those who are using custom WP CRON dropins or setups.

kucrut:
Fixes #1013.

# Checklist

## Release Changelog

- New: Introduce new filter `wp_stream_is_cron_enabled` to allow plugins to override cron status.

## Release Checklist

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -121,7 +121,16 @@ function wp_stream_is_vip() {
  * @return bool
  */
 function wp_stream_is_cron_enabled() {
-	return ( defined( 'DISABLE_WP_CRON' ) && DISABLE_WP_CRON ) ? false : true;
+	$is_cron_enabled = ( defined( 'DISABLE_WP_CRON' ) && DISABLE_WP_CRON ) ? false : true;
+
+	/**
+	 * Filter cron check
+	 *
+	 * @param bool $is_cron_enabled Whether cron is enabled or not.
+	 */
+	$is_cron_enabled = apply_filters( 'wp_stream_is_cron_enabled', $is_cron_enabled );
+
+	return $is_cron_enabled;
 }
 
 /**


includes/functions.php
```diff
@@ -121,7 +121,16 @@ function wp_stream_is_vip() {
...
+	$is_cron_enabled = ( defined( 'DISABLE_WP_CRON' ) && DISABLE_WP_CRON ) ? false : true;
+
+	/**
+	 * Filter cron check
+	 *
+	 * @param bool $is_cron_enabled Whether cron is enabled or not.
+	 */
+	$is_cron_enabled = apply_filters( 'wp_stream_is_cron_enabled', $is_cron_enabled );
+
+	return $is_cron_enabled;
```
kasparsd: @kucrut Could we cast this to boolean to ensure a predictable return value? Also, `$is_cron_enabled` isn't used anywhere apart from storing the filter value so maybe we can return the filtered value directly?

```php
return (bool) apply_filters( 'wp_stream_is_cron_enabled', $is_cron_enabled );
```

kasparsd:
@kucrut I just reviewed #959 and it makes a lot of sense -- the `DISABLE_WP_CRON` applies only to the default WP front-end request triggered CRON events so we shouldn't be checking then when deciding if cron tracking should be allowed.

Let's do #1029 instead. Thanks for opening up this discussion!

PR #1036, opened 2019-09-26
Fix IP exclude rules settings

### LINKED ISSUE #1035: IP addresses in exlude rules not saved correctly
# Bug Report

## Expected Behavior

Exclude rules should be saved correctly, even if some of them do not contain IP addresses.

## Actual Behavior

If there are multiple exclude rules and some of them have no IP address(es) set, settings are not saved correctly. The set IP addresses are then assigned to the wrong authors or roles, because they are displayed one after the other, ignoring whether an IP address was set in the current row at all.

1. Create multiple exclude rules
2. Set IP addresses only for some rules
3. Save settings

## Screenshots
Before save

screenshot-blog bernina local-2019 09 24-13-45-49

After save

screenshot-blog bernina local-2019 09 24-13-46-38

## System Information

- Stream plugin version: [`dev-fix/ip-address`](https://github.com/xwp/stream/tree/fix/ip-address)
- WordPress version: 4.9.11
- PHP version: 7.2.0
- Browser: Google Chrome 77.0
- Computer operating system: macOS Mojave

kasparsd:
Fixes #1035.

HTML `select` elements with nothings selected don't get added to the POST request so the exclude conditions with empty IP rules get moved up to previous rules if they don't have an IP address rule. 

# Checklist


ui/js/exclude.js
```diff
@@ -307,25 +310,12 @@ jQuery(
 
 		$( '#exclude_rules_new_rule' ).on(
 			'click', function() {
-				var $excludeList = $( 'table.stream-exclude-list' );
-
-				$( 'tr:not(.hidden) select.select2-select', $excludeList ).each(
-					function() {
-						$( this ).select2( 'destroy' );
```
kasparsd: Don't recreate the select2 fields for rows with them already present.

ui/js/exclude.js
```diff
@@ -369,6 +359,12 @@ jQuery(
 				$( '.stream-exclude-list tbody tr:not(.hidden) select.select2-select.ip_address', this ).each(
 					function() {
 						var firstSelected = $( 'option:selected', this ).first();
+
+						// Ugly hack to ensure we always pass an empty value or the order of rows gets messed up.
+						if ( ! firstSelected.length ) {
+							$( this ).append( '<option selected="selected"></option>' );
```
kasparsd: This is the core fix -- we add a dummy option and mark it selected to ensure it gets delivered to the backend.

PR #1074, opened 2020-03-20
Store build releases in a distribution repository

### LINKED ISSUE #1073: Composer packages still missing built assets
The Composer package [available on Packagist.org](https://packagist.org/packages/xwp/stream) syncs with this GitHub repository and it uses the default source zip bundle instead of our custom release package.

One solution is to use `wpackagist-plugin/stream` from the [WordPress packagist](https://wpackagist.org/search?q=stream).

Another solution is to introduce a custom distribution repository that contains just the built artifacts similar to how Yoast SEO [does it](https://github.com/Yoast/wordpress-seo/blob/d797de4dedef6f6ce48de9fcc85ca72953239ba8/scripts/deploy_to_dist.sh) with https://github.com/Yoast-dist/wordpress-seo/

kasparsd:
Fixes #1073.

- Store built artifacts in a dedicated repository https://github.com/xwp/stream-dist
- Update the deploy logic to deploy to that repository instead.


kasparsd:
@kopepasah This is now working -- here is a bundle being pushed to the xwp/stream-dist repository with a matching branch https://github.com/xwp/stream-dist/commit/26d58e9e0afa3bb676fc216cfde4a45274edec3f

Could you please review this?

.travis.yml
```diff
@@ -20,11 +20,13 @@ jobs:
       env:
         WORDPRESS_IMAGE_VERSION: php7.2
 
+    - name: Test Release Bundle
+      script: npm run release
```
kasparsd, line 24: This is a standalone script to test the packaging scripts which would previously only run during the tag build which is way too late to fail.

.travis.yml
```diff
@@ -34,6 +36,23 @@ jobs:
...
           - stream-$TRAVIS_TAG.zip
+        on:
+          all_branches: true
+          repo: xwp/stream
+
+    - name: Tag Package
+      if: branch IS present AND type = push
+      script: npm run release
+      before_deploy:
+        - openssl aes-256-cbc -K $encrypted_49634e5b1863_key -iv $encrypted_49634e5b1863_iv -in ./local/travis/travis_deploy_key.enc -out ~/.ssh/id_rsa -d
```
kasparsd, line 47: This is necessary because Travis [doesn't support custom SSH keys for public repos](https://docs.travis-ci.com/user/private-dependencies/#user-key) so we need to manually add a custom SSH key associated with the @team-atlas-bot machine user which has the necessary write permissions.

Gruntfile.js
```diff
@@ -65,6 +65,7 @@ module.exports = function( grunt ) {
 					'languages/*',
 					'readme.txt',
 					'readme.md',
+					'composer.json',
```
kasparsd, line 68: Composer packages must have this in the root.

local/scripts/dist.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
```
kasparsd, line 1: This is where we take the release bundle and push it to the `stream-dist` repository. We always push the branch and add a tag if this is a tag build on travis as indicated by `$TRAVIS_TAG`.

PR #1076, opened 2020-03-25
Fix: Remove old records

### LINKED ISSUE #1056: Old records are not being removed

Duplicate reports:

kasparsd:
Fixes #1056.

- Remove additional location checks to ensure cron tasks can execute.

# Checklist

### DIFF
--- a/classes/class-admin.php
+++ b/classes/class-admin.php
@@ -125,11 +125,6 @@ class Admin {
 
 		add_action( 'init', array( $this, 'init' ) );
 
-		// Ensure function used in various methods is pre-loaded.
-		if ( ! function_exists( 'is_plugin_active_for_network' ) ) {
-			require_once ABSPATH . '/wp-admin/includes/plugin.php';
-		}
-
 		// User and role caps.
 		add_filter( 'user_has_cap', array( $this, 'filter_user_caps' ), 10, 4 );
 		add_filter( 'role_has_cap', array( $this, 'filter_role_caps' ), 10, 3 );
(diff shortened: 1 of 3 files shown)


classes/class-admin.php
```diff
@@ -686,36 +681,17 @@ public function purge_schedule_setup() {
 	public function purge_scheduled_action() {
 		global $wpdb;
 
-		// Don't purge when in Network Admin unless Stream is network activated
-		if (
-			is_multisite()
-			&&
-			is_network_admin()
```
kasparsd: I'm not sure why `is_network_admin()` would need to be checked on cron runs.

classes/class-settings.php
```diff
@@ -270,13 +270,7 @@ public function add_display_name_search_columns( $search_columns, $search, $quer
 	public function get_option_key() {
 		$option_key = $this->option_key;
 
-		$current_page = wp_stream_filter_input( INPUT_GET, 'page' );
-
-		if ( ! $current_page ) {
-			$current_page = wp_stream_filter_input( INPUT_GET, 'action' );
-		}
-
-		if ( 'wp_stream_network_settings' === $current_page ) {
```
kasparsd: Use the network settings if the plugin is network-activated. Is there a situation we would like to use network settings when the plugin is _not_ network activated?

kasparsd:
It has been a while since last looked into this and it would be great if we could fix the actual issue without changing so much of the surrounding code. I'm going to close this for now and we should create a new pull request just for the bugfix.

PR #1086, opened 2020-04-28
Fix/foreign lang connector bug

### LINKED ISSUE #1051: Connector "blogs" is added to all searches on a multisite if any sub-sites has a non-english site title
#### Bug Report

If anyone of the subsites in a network has a non-english site title, Clicking "Search Records" in stream records auto-selects "blogs" as a connector.

Here is a video demonstrating the bug - https://share.getcloudapp.com/d5u0l5YE

#### Expected Behavior
Clicking "Search Records" should not auto-select a connector as that limits the actual search results.

**Here is what I found after some debugging -** 

- Stream, when generating a list of all the sites uses function `sanitize_key()` in [blogs](https://github.com/xwp/stream/blob/31ae9b9362d303a55bd6b5e15cf0ad6a3ec10620/connectors/class-connector-blogs.php#L77) and [mercator](https://github.com/xwp/stream/blob/143d98dae53054b82bbafd406ab2dde3d4952719/connectors/class-connector-mercator.php#L68) connectors, over this non-english text and it is returned as `''`.
- When building the select fields for the connectors, the select field value gets selected from [selected()](https://github.com/xwp/stream/blob/fac60647d704fb7042b2969d3d978c8486219871/classes/class-list-table.php#L750)
- This selected value is then applied to the select field from [admin.js](https://github.com/xwp/stream/blob/fac60647d704fb7042b2969d3d978c8486219871/ui/js/admin.js#L135)

oscarssanchez:
Fixes #1051.

Hi @kasparsd ,

Could you use this PR that solves #1051 ?

This simply removes `sanitize_key()`. It looks like this value is later escaped:
https://github.com/xwp/stream/blob/develop/classes/class-list-table.php#L780 as well as label: https://github.com/xwp/stream/blob/develop/classes/class-list-table.php#L787

Since this is not getting saved into the database, my idea is it won't do any harm removing it, and there's no benefit adding an escaping function, which was my other solution. 

The unit tests seem to pass, as well as the problem is fixed by testing locally.

Thanks!

# Checklist

## Release Changelog

## Release Checklist

### DIFF
--- a/connectors/class-connector-blogs.php
+++ b/connectors/class-connector-blogs.php
@@ -74,7 +74,7 @@ class Connector_Blogs extends Connector {
 
 			foreach ( $blogs as $blog ) {
 				$blog_details   = get_blog_details( $blog->blog_id );
-				$key            = sanitize_key( $blog_details->blogname );
+				$key            = $blog_details->blogname;
 				$labels[ $key ] = $blog_details->blogname;
 			}
 		}
--- a/connectors/class-connector-blogs.php
+++ b/connectors/class-connector-blogs.php
@@ -74,7 +74,7 @@ class Connector_Blogs extends Connector {
 
 			foreach ( $blogs as $blog ) {
 				$blog_details   = get_blog_details( $blog->blog_id );
-				$key            = $blog_details->blogname;
+				$key            = sprintf( 'blog-%d', $blog->blog_id );
 				$labels[ $key ] = $blog_details->blogname;
 			}
 		}
--- a/connectors/class-connector-mercator.php
+++ b/connectors/class-connector-mercator.php
@@ -65,7 +65,7 @@ class Connector_Mercator extends Connector {
 
 			foreach ( $blogs as $blog ) {
 				$blog_details   = get_site( $blog->blog_id );
-				$key            = $blog_details->blogname;
+				$key            = sprintf( 'blog-%d', $blog->blog_id );
 				$labels[ $key ] = $blog_details->blogname;
 			}
 		}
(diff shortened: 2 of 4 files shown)


connectors/class-connector-blogs.php
```diff
@@ -74,7 +74,7 @@ public function get_context_labels() {
 
 			foreach ( $blogs as $blog ) {
 				$blog_details   = get_blog_details( $blog->blog_id );
-				$key            = sanitize_key( $blog_details->blogname );
```
kasparsd, line 77: Looks like context label keys have to be globally unique so how about we used something like `sprintf( 'blog-%d', $blog->blog_id );` for the key? That should be safe in every instance.

Would this break any of the existing records and the way they're filtered?

kasparsd:
@oscarssanchez Could you please also review the failing unit tests on single sites https://travis-ci.com/github/xwp/stream/jobs/324102448#L573? Thanks!

connectors/class-connector-blogs.php
oscarssanchez: You are right @kasparsd `blog_id` would be safer to use rather than `blog_name` !

oscarssanchez: This didn't break existing records on my testing.

oscarssanchez:
Hi @kasparsd ,

Thanks for your feedback, I really appreciate you taking the time to review this, and sorry for the delay, as it has been busy weeks. 

I have updated the PR with your suggestion and updated the Unit Tests to reflect them. 

Thanks again!

PR #1104, opened 2020-06-11
Fixes broken date in MySQL 8

### LINKED ISSUE #1091: Incorrect Date and Time
See https://wordpress.org/support/topic/incorrect-date-and-time/

> I’m having an issue with the plugin, that the date and time for all records is being set as
> 10 years ago
> -0001/11/30
> 12:00:00 AM

kidunot89:
Fixes #1091 

Replaces old `sprintf( ... )` formatting logic with `$date->format( DateTime::ATOM )`  to create the ISO 8601 formatted timestamp to be used in the record.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,7 +20,7 @@ services:
       - mysql
       - db_phpunit
     ports:
-      - "8080:80"
+      - "80:80"
       - "443:443"
       - "9000:9000" # Xdebug
     volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,7 @@ version: '3'
 services:
 
   mysql:
-    image: mysql:8
+    image: mysql:5
     command: '--default-authentication-plugin=mysql_native_password'
     volumes:
       - db_data:/var/lib/mysql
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -3,7 +3,8 @@ version: '3'
 services:
 
   mysql:
-    image: mysql:5
+    image: mysql:8
+    command: '--default-authentication-plugin=mysql_native_password'
```
kasparsd, line 7: What is the purpose of this argument? Is it needed because of the version bump from 5 to 8?

includes/functions.php
```diff
@@ -53,12 +53,7 @@ function wp_stream_get_iso_8601_extended_date( $time = false, $offset = 0 ) {
 	$timezone = new DateTimeZone( $offset_string );
 	$date     = new DateTime( gmdate( 'Y-m-d H:i:s.' . $micro_seconds, $microtime ), $timezone );
 
-	return sprintf(
-		'%s%03d%s',
-		$date->format( 'Y-m-d\TH:i:s.' ),
-		floor( $date->format( 'u' ) / 1000 ),
-		$date->format( 'O' )
-	);
+	return $date->format( DateTime::ATOM );
```
kasparsd, line 56: This looks like a simple fix! Could we add a unit test for this?

Also, were you able to confirm that `strtotime( $record->created )` is converted to a timestamp correctly?

https://github.com/xwp/stream/blob/49fea4ab4b6659356115c967af981e027e7b4f63/classes/class-list-table.php#L266

docker-compose.yml
```diff
@@ -19,7 +20,7 @@ services:
       - mysql
       - db_phpunit
     ports:
-      - "80:80"
+      - "8080:80"
```
kasparsd, line 23: Port 80 shouldn't cause conflicts with existing services when running inside the Vagrant wrapper since it uses a dedicated network interface. The documentation and instructions still use the Vagrant wrapper so maybe we can adjust this later if needed.

kidunot89:
@kasparsd Unit test added

kasparsd:
Opening a new pull request since I'm unable to make updates to the current source branch.

PR #1111, opened 2020-06-15
WordPress VIP coding standards compliance

kidunot89:
# Summary
Adds the VIP coding standards and clears up all the warnings from the PHPCS test.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "phpcs.standard": "phpcs.xml.dist"
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "phpcs.standard": "phpcs.xml.dist"
-}
--- a/classes/class-admin.php
+++ b/classes/class-admin.php
@@ -479,13 +479,17 @@ class Admin {
 				)
 			);
 
+			$order_types = array( 'asc', 'desc' );
+
 			wp_localize_script(
 				'wp-stream-live-updates',
 				'wp_stream_live_updates',
 				array(
 					'current_screen'      => $hook,
-					'current_page'        => isset( $_GET['paged'] ) ? esc_js( $_GET['paged'] ) : '1', // phpcs:ignore WordPress.Security.NonceVerification.Recommended
-					'current_order'       => isset( $_GET['order'] ) ? esc_js( $_GET['order'] ) : 'desc', // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+					'current_page'        => isset( $_GET['paged'] ) ? absint( wp_unslash( $_GET['paged'] ) ) : '1', // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+					'current_order'       => isset( $_GET['order'] ) && in_array( strtolower( $_GET['order'] ), $order_types, true ) // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+						? esc_js( $_GET['order'] ) // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+						: 'desc',
 					'current_query'       => wp_stream_json_encode( $_GET ), // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 					'current_query_count' => count( $_GET ), // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 				)
--- a/classes/class-admin.php
+++ b/classes/class-admin.php
@@ -254,7 +254,7 @@ class Admin {
 	 */
 	public function notice( $message, $is_error = true ) {
 		if ( defined( 'WP_CLI' ) && WP_CLI ) {
-			$message = strip_tags( $message );
+			$message = wp_strip_all_tags( $message );
 
 			if ( $is_error ) {
 				WP_CLI::warning( $message );
@@ -392,7 +392,7 @@ class Admin {
 	 *
 	 * @action admin_enqueue_scripts
 	 *
-	 * @param string $hook
+	 * @param string $hook  Current hook.
 	 *
 	 * @return void
 	 */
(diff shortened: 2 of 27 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "phpcs.standard": "phpcs.xml.dist"
```
kasparsd, line 2: Is this required for one of the `PHPCS` extensions for VS Code? Both `phpcs.xml` and `phpcs.xml.dist` are some of the default filenames resolved by the phpcs runner:

> If you run PHP_CodeSniffer without specifying a coding standard, PHP_CodeSniffer will look in the current directory, and all parent directories, for a file called either .phpcs.xml, phpcs.xml, .phpcs.xml.dist, or phpcs.xml.dist

https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#using-a-default-configuration-file

and:

> phpcs.autoConfigSearch [ Scope: All | Optional | Type: boolean | Default: true ]
>
> Automatically search for any phpcs.xml, phpcs.xml.dist, phpcs.ruleset.xml or ruleset.xml file to use as configuration. Overrides phpcs.standard configuration when a ruleset is found.

https://marketplace.visualstudio.com/items?itemName=ikappas.phpcs

classes/class-admin.php
```diff
@@ -469,14 +484,10 @@ public function admin_enqueue_scripts( $hook ) {
...
 					'current_screen'      => $hook,
-					'current_page'        => isset( $_GET['paged'] ) ? esc_js( $_GET['paged'] ) : '1', // WPCS: CSRF ok.
-					// input var okay, CSRF okay
-					'current_order'       => isset( $_GET['order'] ) ? esc_js( $_GET['order'] ) : 'desc', // WPCS: CSRF ok.
-					// input var okay, CSRF okay
-					'current_query'       => wp_stream_json_encode( $_GET ), // WPCS: CSRF ok.
-					// input var okay, CSRF okay
-					'current_query_count' => count( $_GET ), // WPCS: CSRF ok.
-					// input var okay, CSRF okay
+					'current_page'        => isset( $_GET['paged'] ) ? esc_js( $_GET['paged'] ) : '1', // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
kasparsd, line 487: Should this be sanitized to an integer value?

classes/class-admin.php
```diff
@@ -469,14 +484,10 @@ public function admin_enqueue_scripts( $hook ) {
...
+					'current_order'       => isset( $_GET['order'] ) ? esc_js( $_GET['order'] ) : 'desc', // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
kasparsd, line 488: Could we check that it's one of the allowed values of `order` before setting it? Do we support anything apart from `desc` and `asc`?

classes/class-admin.php
```diff
@@ -469,14 +484,10 @@ public function admin_enqueue_scripts( $hook ) {
...
+					'current_query'       => wp_stream_json_encode( $_GET ), // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
kasparsd, line 489: This appears to be passing the whole `$_GET` to the page DOM. This has a lot of potential for abuse.

composer.json
```diff
@@ -14,6 +14,7 @@
     "composer/installers": "~1.0"
   },
   "require-dev": {
+    "automattic/vipwpcs": "@stable",
```
kasparsd, line 17: Could we add the version following the SEMVER syntax similar to the rest of the dependencies. This would also ensure that Dependabot can open pull requests for dependency upgrade and we're keeping the same style for clarify.

classes/class-admin.php
derekherman: `absint( wp_unslash( $_GET['paged] ) )` would probably be my preference here, but there are many ways to sanitize/escape this.

classes/class-admin.php
derekherman: I agree, if there are only `asc` and `desc` we can do a check for one and manually set this instead of excepting any value.

classes/class-admin.php
derekherman: Yeah these are scary

derekherman: Though to be fair are not a change and should probably be addressed in a new PR

composer.json
derekherman: `"^2.0.0"` should do it

local/config/wp-tests-config.php
```diff
@@ -3,6 +3,7 @@
  * Config file used only for phpunit tests inside Docker.
  *
  * phpcs:disable WordPress.WP.GlobalVariablesOverride.Prohibited
+ * phpcs:disable Squiz.Commenting.FileComment.MissingPackageTag
```
derekherman, line 6: Why not just add the package name?

.vscode/settings.json
kidunot89: Naw, I've just always set my workspace setting up like that. I forgot to add it to the `.gitignore`

classes/class-admin.php
kidunot89: I gotcha, I wasn't worried about the code and was just trying to silence the PHPCS warning.

local/config/wp-tests-config.php
kidunot89: Because this is a `wp-config.php` file and won't live within the plugin during use. :man_shrugging:

classes/class-log.php
```diff
@@ -154,7 +154,7 @@ function ( $var ) {
 		$result = $this->plugin->db->insert( $recordarr );
 
 		// This is helpful in development environments:
-		// error_log( $this->debug_backtrace( $recordarr ) );
+		// error_log( $this->debug_backtrace( $recordarr ) );.
```
kasparsd, line 157: Could we remove this completely?

kasparsd:
Looks like need to resolve the merge conflict. @kidunot89 could you do that, please?

phpcs.xml.dist
```diff
@@ -23,6 +23,6 @@
 	<exclude-pattern>*/ui/lib/*</exclude-pattern>
 	<exclude-pattern>*/vendor/*</exclude-pattern>
 	<exclude-pattern>*/build/*</exclude-pattern>
-	<exclude-pattern>*/local/public/*</exclude-pattern>
+	<exclude-pattern>*/local/*</exclude-pattern>
```
kidunot89, line 26: Since it only used is in development, I've added the whole `local` directory to be ignored by PHPCS.

kidunot89: @kasparsd @derekherman :point_up:

kasparsd: It would be nice to have all PHP files with consistent formatting even if they're used for development purposes.

PR #1113, opened 2020-06-16
Allow filtering the ignored options.

### LINKED ISSUE #1078: Support ignoring certain options or option keys
# Feature Request

Stream has a [concept of ignored options](https://github.com/xwp/stream/blob/88317c6149f9a75d53db95c2cf5f10654a4b5a71/connectors/class-connector-settings.php#L338) but the list of options to ignore is static.

Some options are updated frequently by an automated fashion, not in response to a user action. This can lead to cluttered stream logs that aren't relevant to the end user.

It'd be helpful if a developer could exclude those options from being logged by stream. I think a filter on the `$ignored` array in both `is_key_ignored` and `is_option_ignored` would work well.

kopepasah:
Fixes #1078.

Adds a filter for allowing/disallowing options.

# Checklist

### DIFF
--- a/connectors/class-connector-settings.php
+++ b/connectors/class-connector-settings.php
@@ -344,13 +344,27 @@ class Connector_Settings extends Connector {
 			return true;
 		}
 
-		$ignored = array(
+		$default_ignored = array(
 			'image_default_link_type',
 			'medium_large_size_w',
 			'medium_large_size_h',
 		);
 
-		return in_array( $option_name, $ignored, true );
+		/**
+		 * Filters the boolean output for is_option_ignored().
+		 *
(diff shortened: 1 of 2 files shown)


kopepasah:
@kasparsd currently there are no Unit Tests for any of the Connectors. Just adding that note, as we may want to tackle all those at once.

kasparsd:
Feel free to merge if you think we should add the unit test in a separate pull request.

kopepasah:
@kasparsd I went ahead and added a test (please review). 

What I meant previously was we could tackle this when we tackle tests as a whole, but it makes more sense to add tests as we go too. ;)

PR #1114, opened 2020-06-16
Update `readme` for w.org plugins page to match new copy

postphotos:
Updated readme to match new copy that more accurately describes Stream.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -7,18 +7,20 @@ Stable tag: 3.4.3
 License: GPLv2 or later
 License URI: https://www.gnu.org/licenses/gpl-2.0.html
 
-Planes have a black box, WordPress has Stream. When something goes wrong, you need to know how it happened.
-
+With Stream, you’re never left in the dark about changes to your WordPress site.
 
 == Description ==
 
-With Stream, you're never left in the dark about WordPress Admin activity.
+With real-time notifications and third-party integrations, Stream can proactively alert you when something goes wrong with your WordPress site. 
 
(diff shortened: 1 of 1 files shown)


readme.txt
```diff
@@ -1,24 +1,26 @@
 === Stream ===
-Contributors: lukecarbis, fjarrett, stream, xwp, kasparsd
+Contributors: xwp, kasparsd, derekherman, postphotos
```
derekherman, line 2: This should probably just be `xwp`. Not sure who has access to the stream account on wp.org and this list gives everyone write access to SVN. @kasparsd I can either give you the credentials to the `xwp` account so you can still push new versions of the plugin or keep you on the list but I don't see why I or Leo would be on this list. Thoughts?

postphotos: Agreed. Just pushed a change.

postphotos:
Hey @derekherman - I've added the list of contributors as @kasparsd suggested and removed the other non-repo-committing folks from the list.

I think this is good to merge now.

readme.txt
kasparsd: > Not sure who has access to the stream account on wp.org and this list gives everyone write access to SVN

@derekherman This contributors list is only included in the plugin credits and doesn't give SVN access. 

contributors

This attribution also makes the plugin show up on [user's list of contributions](https://profiles.wordpress.org/kasparsd/#content-plugins) which is nice:

contributions

The SVN access is managed via Committers under [the plugin settings](https://wordpress.org/plugins/stream/advanced/):

committers

Issue #1093, opened 2020-05-28
Improve Test Coverage


kidunot89:
# Feature Request

The alert and connector class have little to now coverage.
- Alert types
  - ![image](https://user-images.githubusercontent.com/13604318/83191700-6f727500-a102-11ea-817c-8d49201aa386.png)
- Connectors
  - ![image](https://user-images.githubusercontent.com/13604318/83191612-510c7980-a102-11ea-96bd-afdc5d7b76bb.png)

The plugin core classes have better coverage, but still aren't that good.
![image](https://user-images.githubusercontent.com/13604318/83191992-e445af00-a102-11ea-9ebb-21db01cb0fd7.png)
_(Sorry about the artifact :man_facepalming:)_

kasparsd:
@kidunot89 Would you like to keep this issue open until we're reached a certain test coverage range? Or can we close it in favour of individual tickets for specific parts of code?

kidunot89:
@kasparsd Let's leave this open until Code coverage is at an acceptable level. I also want to discuss what to do about parts of the code where simulation isn't really an option, like WP core update logging functionality.

kasparsd:
Looks like this keeps getting closed because of the `Fixes #....` references in the pull requests @kidunot89 :)

kidunot89:
:man_facepalming:

PR #1118, opened 2020-06-23
ACF connector test implemented and ACF v5.x support added

### LINKED ISSUE #1122: Support for Advanced Custom Fields (ACF) version 5
> When updating an ACF repeater field (which is inside an ACF Group field),
the Stream plugin crashes as the mysql escape function from WP_DB gets incorrect parameters passed.

https://wordpress.org/support/topic/error-crash-when-updating-acf-repeater-field/

kidunot89:
Partially fixes #1093 
Fixes and closes #1122 

# Summary
- Updates the ACF connector class to work with the changes made in ACF v5+
- Adds tests for ACF v5+ connector logic.
- Adds `wpackagist` repository and `wpackagist-plugin/advanced-custom-fields` dev dependency to the `composer.json`.
- Updates tests `bootstrap` file to activate plugins defined in the PHPUnit const `WP_TEST_ACTIVATED_PLUGINS`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,6 +4,12 @@
   "homepage": "https://wordpress.org/plugins/stream/",
   "type": "wordpress-plugin",
   "license": "GPL-2.0-or-later",
+  "repositories": [
+    {
+      "type":"composer",
+      "url":"https://wpackagist.org"
+    }
+  ],
   "require": {
     "composer/installers": "~1.0"
   },
@@ -15,7 +21,8 @@
     "wp-cli/wp-cli-bundle": "^2.2",
     "wp-coding-standards/wpcs": "^2.2",
     "wp-phpunit/wp-phpunit": "^5.4",
-    "wpsh/local": "^0.2.3"
+    "wpsh/local": "^0.2.3",
+    "wpackagist-plugin/advanced-custom-fields": "5.8.12"
   },
   "config": {
     "process-timeout": 600,
@@ -25,7 +32,10 @@
     }
   },
   "extra": {
-    "wordpress-install-dir": "local/public"
+    "wordpress-install-dir": "local/public",
+    "installer-paths": {
+      "local/public/wp-content/plugins/{$name}/": ["type:wordpress-plugin"]
+    }
   },
   "scripts": {
     "release": [
--- a/connectors/class-connector-acf.php
+++ b/connectors/class-connector-acf.php
@@ -31,6 +31,8 @@ class Connector_ACF extends Connector {
 	 * @var array
 	 */
 	public $actions = array(
+		'save_post',
+		'post_updated',
 		'added_post_meta',
 		'updated_post_meta',
 		'delete_post_meta',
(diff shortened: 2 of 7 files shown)


composer.json
```diff
@@ -19,6 +23,7 @@
     "wordpress/wordpress": "^5.2",
     "wp-cli/wp-cli-bundle": "^2.2",
     "wp-coding-standards/wpcs": "^2.2",
+    "wpackagist-plugin/advanced-custom-fields": "*",
```
kopepasah, line 26: @kidunot89 forgive my popping in without context, but what is the reason for requiring ACF? This believe will install this plugin in cases someone manages Stream installations using Composer.

kidunot89: :thinking: See as this is a **dev** dependency, I don't believe that is the case.

kopepasah: @kidunot89 whoops. I did not expand the view and mistakenly thought this was under `require`. 😬  I will not install as require-dev is read a `root` only.

connectors/class-connector-acf.php
```diff
@@ -147,13 +149,219 @@ public function action_links( $links, $record ) {
...
+	}
+
 	/**
 	 * Track addition of post meta
 	 *
 	 * @action added_post_meta
 	 */
 	public function callback_added_post_meta() {
-		call_user_func_array( array( $this, 'check_meta' ), array_merge( array( 'post', 'added' ), func_get_args() ) );
+		$this->check_meta( 'post', 'added', ...func_get_args() );
```
kopepasah, line 364: @kidunot89 what is the PHP version compatibility of the spread operator?

kopepasah:
@kidunot89 my only concern here is requiring the ACF plugin from Composer for development. If we were to follow this same practice, we could end up with a ton of WordPress plugins we are adding the the Composer config, which are not really required for development, but rather for testing. Installing of these plugins for the PHPUnit tests may better be served in some other scripted process for testing, but would like to hear your thoughts on why this method is prefered.

kidunot89:
@kopepasah Yes, My plan is to add all the plugins, supported by stream through connectors, to the `dev` dependencies, and writing an integration test for each connector. My reasoning is simple, is the best option for testing the **connectors**.

connectors/class-connector-acf.php
kidunot89: :man_facepalming: 7.4, my bad :sweat_smile:.

kidunot89: @kopepasah Upon doing further research, I learned that this is called **Argument Unpacking** in PHP, and yes it was implemented in PHP 5.6. Official RFC [here](https://wiki.php.net/rfc/argument_unpacking)

kasparsd:
@kidunot89 I suspect the failing unit tests are caused by wp-dev-lib copying over the project source files to a different place for the actual test run:

https://github.com/xwp/wp-dev-lib/blob/420f1bac828e105a19a904cab97d042891d4b84e/scripts/check-diff.sh#L757-L765

So none of the dependencies (plugins) are moved over to the same `wp-content/plugins` directory. This is a great example of the many assumptions wp-dev-lib makes about the project structure.

I guess at this point we can switch to using the Docker test environment on Travis too. There are ways to cache the Docker images once they're built so we should be fine.

connectors/class-connector-acf.php
kopepasah: Thanks for the additional research! I was not aware it had been around for so long. 😱

kidunot89:
@kasparsd I don't believe that to be the case due to the fact that the same configurations are being used in the bbPress connector PR and CI is passing :point_right: #1120

kasparsd:
@kidunot89 Unfortunately this is another issue with wp-dev-lib -- those tests are passing because it didn't even run the tests https://travis-ci.com/github/xwp/stream/jobs/355627134#L490 -- the `source "$DEV_LIB_PATH/travis.script.sh"` command failed silently...

kopepasah:
@kidunot89 just a quick question, regarding ACF v5. Do we need to perform any version checks for backwards compatibility or does the functionality simply not exist pre-v5?

kidunot89:
@kopepasah Pre-v5 functionality exists, it's just not tested, or updated.

connectors/class-connector-acf.php
```diff
@@ -147,13 +149,210 @@ public function action_links( $links, $record ) {
...
+		// Log new ACF field additions to field groups.
+		if ( 'acf-field' === $post->post_type ) {
+			$parent = get_post( $post->post_parent );
+			if ( $parent ) {
+				$this->log_prop( 'added', $post_id, $post, 'parent', $parent );
+			}
+		} elseif ( 'acf-field-group' === $post->post_type ) {
+			$props = maybe_unserialize( $post->post_content );
+
+			if ( ! empty( $props ) ) {
```
kopepasah, line 176: @kidunot89 `maybe_unserialize()` does not always return an array. Do you think we need to check that `$props` is an array before running the `foreach` loop?

connectors/class-connector-acf.php
```diff
@@ -479,10 +678,31 @@ public function log_override( $data ) {
 			return $data;
 		}
 
-		if ( 'posts' === $data['connector'] && 'acf' === $data['context'] ) {
-			$data['context']               = 'field_groups';
-			$data['connector']             = $this->name;
-			$data['args']['singular_name'] = esc_html__( 'field group', 'stream' );
+		$is_acf_context = 'acf' === $data['context']
```
kopepasah, line 681: @kidunot89 how about this: 

```php
$is_acf_context = in_array( $data['context'], array( 'acf', 'acf-field-group', 'acf-field' ), true );
```

tests/tests/connectors/test-class-connector-acf.php
```diff
@@ -0,0 +1,193 @@
+<?php
+namespace WP_Stream;
```
kopepasah, line 2: I am not certain how @kasparsd feels, but I would like to see some more documentation in these tests.

tests/tests/connectors/test-class-connector-acf.php
```diff
@@ -0,0 +1,193 @@
...
+		$this->mock = $this->getMockBuilder( Connector_ACF::class )
+			->setMethods( [ 'log' ] )
+			->getMock();
+
+		// Register connector.
+		$this->mock->register();
+
+		// Reusable ACF object keys.
+		$this->group_key = __CLASS__;
+		$this->field_key = 'field_5d7812fd000a4';
```
kopepasah, line 22: @kidunot89 is this field key unique (always expected to be this value) or arbitrary?

tests/tests/connectors/test-class-connector-acf.php
```diff
@@ -0,0 +1,193 @@
...
+		$defaults = [
+			'parent'            => $this->group_key,
+			'key'               => $this->field_key,
+			'label'             => 'Test Field',
+			'name'              => 'test_field',
+			'type'              => 'text',
+			'instructions'      => '',
+			'required'          => 0,
+			'conditional_logic' => 0,
+			'wrapper'           => array(
```
kopepasah, line 66: Sometimes you are using shorthand arrays and other times using the long form of array. We need to pick one an stick with that form throughout the plugin.

kopepasah:
> @kopepasah Pre-v5 functionality exists, it's just not tested, or updated.

Thanks for that information! It is difficult to tell, just by reviewing, which parts are pre-v5 and remain backwards compatible, but as long as you are aware, then that can be resolved when testing the changes.

I added my review and requested just a few changes.

connectors/class-connector-acf.php
kidunot89: :man_facepalming: Can't believe I didn't think of that :joy:

tests/tests/connectors/test-class-connector-acf.php
kidunot89: There was a doc block, but it was lost in PR somewhere. I'll correct this.

tests/tests/connectors/test-class-connector-acf.php
kidunot89: No, I was using this in an old iteration the test, I remove it.

tests/tests/connectors/test-class-connector-acf.php
kidunot89: For now, I go with long version because it used through the plugin.

kidunot89:
@kopepasah I add notes in the doc blocks for the pre ACF-v5 callbacks.

tests/tests/connectors/test-class-connector-acf.php
kidunot89: This has been [fixed](https://github.com/xwp/stream/pull/1118/files#diff-8b4d036afce8fdd07963c1eba24744a1R2-R6), but the snippet isn't updated :thinking:

kasparsd:
This looks good @kidunot89! Feel free to merge if there is nothing to add.

PR #1128, opened 2020-07-07
Fix CI build times

kasparsd:
- [x] Use Docker Hub to host the WordPress images https://hub.docker.com/repository/docker/xwpco/stream-wordpress

- [x] Add tooling to build and push images to https://hub.docker.com/repository/docker/xwpco/stream-wordpress

- [x] Update Travis config to use those images instead of building them or reading them from file cache which was taking a long time with larger images.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,6 @@ services:
   - docker
 
 before_install:
-  - docker-compose --file docker-compose.build.yml build
   - nvm install
   - nvm use
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,11 @@ jobs:
   include:
 
     - stage: test
-      name: Test with PHP 5.6
+      name: Default Test
+      after_script:
+        - test -f ./tests/reports/clover.xml && composer test-report || true
+
+    - name: Test with PHP 5.6
       env:
         WORDPRESS_IMAGE_VERSION: php5.6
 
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -9,13 +9,14 @@ jobs:
 
     - stage: test
       name: Test with PHP 5.6
-      before_script: docker-compose build --build-arg PHP_VERSION=5.6 --build-arg XDEBUG_VERSION=2.5.5 wordpress
+      env:
+        WORDPRESS_IMAGE_VERSION: php5.6
```
kasparsd, line 13: `docker-compose.yml` now uses the `WORDPRESS_IMAGE_VERSION` environment variable for determining the version of docker image to use.

.travis.yml
```diff
@@ -25,7 +26,7 @@ services:
   - docker
 
 before_install:
-  - docker load -i /tmp/docker_images/images.tar || true
+  - docker-compose pull wordpress
```
kasparsd, line 29: This pulls them in from https://hub.docker.com/repository/docker/xwpco/stream-wordpress

docker-compose.build.yml
```diff
@@ -0,0 +1,23 @@
+version: '3'
```
kasparsd, line 1: Define the images and the build arguments for each tag.

PR #1139, opened 2020-07-10
GravityForms connector test implemented

kidunot89:
Partially fixes #1093 

# Summary checklist
- [x] `gform_after_save_form` callback test implemented and passing.
- [x] `gform_delete_lead` callback test implemented and passing.
- [x] `gform_pre_confirmation_save` callback test implemented and passing.
- [x] `gform_pre_notification_save` callback test implemented and passing.
- [x] `gform_pre_notification_deleted` callback test implemented and passing.
- [x] `gform_pre_confirmation_deleted` callback test implemented and passing.
- [x] `gform_confirmation_status` callback test implemented and passing.
- [x] `gform_notification_status` callback test implemented and passing.
- [x] `check` function test implemented and passing.
- [x] `gform_post_note_added` callback test implemented and passing.
- [x] `gform_pre_note_deleted` callback test implemented and passing.
- [x] `gform_update_status` callback test implemented and passing.
- [x] `gform_update_is_read` callback test implemented and passing.
- [x] `gform_update_is_starred` callback test implemented and passing.
- [x] `log_form_action` function test implemented and passing.
- [x] `gform_post_export_entries` callback test implemented and passing.
- [x] `gform_forms_post_import` callback test implemented and passing.
- [x] `gform_export_separator` callback refactored. It's now the `gform_export_form` callback and has been tested.
- No test was written for the`gform_export_options` callback. Further inspection is needed to determine it's relevance.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,7 @@ jobs:
       if: branch IS present AND type = push
       script: npm run release
       before_deploy:
+        - git restore composer.json
         - openssl aes-256-cbc -K $encrypted_49634e5b1863_key -iv $encrypted_49634e5b1863_iv -in ./local/travis/travis_deploy_key.enc -out ~/.ssh/id_rsa -d
         - chmod 600 ~/.ssh/id_rsa
       deploy:
@@ -71,7 +72,25 @@ before_install:
   - composer self-update --1
 
 install:
+  - |
+    composer config repositories."1"  "{
+      \"type\": \"package\",
+      \"package\": {
+        \"name\": \"gravityforms/gravityforms\",
+        \"version\": \"2.4.21.3\",
+        \"type\": \"wordpress-plugin\",
+        \"dist\": {
+          \"type\": \"zip\",
+          \"url\": \"https://www.gravityhelp.com/wp-content/plugins/gravitymanager/api.php?op=get_plugin&slug=gravityforms&key=$GRAVITYFORMS_KEY\"
+        },
+        \"require\": {
+          \"composer/installers\": \"^1.4\",
+          \"gotoandplay/gravityforms-composer-installer\": \"^2.3\"
+        }
+      }
+    }"
   - npm install
+  - composer require --dev gravityforms/gravityforms:2.4.21.3
 
 script:
   - npm run lint
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,23 @@
   "repositories": [
     {
       "type": "composer",
-      "url": "https://wpackagist.org"
+      "url":"https://wpackagist.org"
+    },
+    {
+      "type": "package",
+      "package": {
+        "name": "gravityforms/gravityforms",
+        "version": "2.4.21.3",
+        "type": "wordpress-plugin",
+        "dist": {
+          "type": "zip",
+          "url": "https://www.gravityhelp.com/wp-content/plugins/gravitymanager/api.php?op=get_plugin&slug=gravityforms&key={%GRAVITYFORMS_KEY}"
+        },
+        "require": {
+          "composer/installers": "^1.4",
+          "gotoandplay/gravityforms-composer-installer": "^2.3"
+        }
+      }
     }
   ],
   "require": {
@@ -25,15 +41,12 @@
     "wpackagist-plugin/advanced-custom-fields": "5.8.12",
     "wpackagist-plugin/easy-digital-downloads": "^2.9.23",
     "wpackagist-plugin/user-switching": "^1.5.5",
-    "wp-premium/gravityforms": "^2.4.17",
+    "gravityforms/gravityforms": "2.4.21.3",
     "wpsh/local": "^0.2.3"
   },
   "config": {
     "process-timeout": 600,
-    "sort-packages": true,
-    "platform": {
-      "php": "5.6.20"
-    }
+    "sort-packages": true
   },
   "extra": {
     "wordpress-install-dir": "local/public",
--- a/composer.json
+++ b/composer.json
@@ -8,22 +8,6 @@
     {
       "type": "composer",
       "url":"https://wpackagist.org"
-    },
-    {
-      "type": "package",
-      "package": {
-        "name": "gravityforms/gravityforms",
-        "version": "2.4.21.3",
-        "type": "wordpress-plugin",
-        "dist": {
-          "type": "zip",
-          "url": "https://www.gravityhelp.com/wp-content/plugins/gravitymanager/api.php?op=get_plugin&slug=gravityforms&key={%GRAVITYFORMS_KEY}"
-        },
-        "require": {
-          "composer/installers": "^1.4",
-          "gotoandplay/gravityforms-composer-installer": "^2.3"
-        }
-      }
     }
   ],
   "require": {
@@ -41,7 +25,6 @@
     "wpackagist-plugin/advanced-custom-fields": "5.8.12",
     "wpackagist-plugin/easy-digital-downloads": "^2.9.23",
     "wpackagist-plugin/user-switching": "^1.5.5",
-    "gravityforms/gravityforms": "2.4.21.3",
     "wpsh/local": "^0.2.3"
   },
   "config": {
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,7 @@
     "wpackagist-plugin/advanced-custom-fields": "5.8.12",
     "wpackagist-plugin/easy-digital-downloads": "^2.9.23",
     "wpackagist-plugin/user-switching": "^1.5.5",
+    "wp-premium/gravityforms": "^2.4.17",
     "wpsh/local": "^0.2.3"
   },
   "config": {
--- a/tests/tests/connectors/test-class-connector-gravityforms.php
+++ b/tests/tests/connectors/test-class-connector-gravityforms.php
@@ -1,0 +1,1054 @@
+<?php
+/**
+ * WP Integration Test w/ GravityForms
+ *
+ * Tests for GravityForms Connector class callbacks.
+ */
+
+namespace WP_Stream;
+
+class Test_WP_Stream_Connector_GravityForms extends WP_StreamTestCase {
+	/**
+	 * Runs before each test
+	 */
+	public function setUp() {
+		parent::setUp();
+
+		$this->plugin->connectors->unload_connectors();
+
+		$post_connector = new Connector_Posts();
+		$post_connector->register();
+
+		// Make partial of Connector_GravityForms class, with mocked "log" function.
+		$this->mock = $this->getMockBuilder( Connector_GravityForms::class )
+			->setMethods( array( 'log' ) )
+			->getMock();
+
+		$this->mock->register();
+	}
+
+	/**
+	 * Runs after each test
+	 */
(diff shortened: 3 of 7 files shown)


tests/bootstrap.php
```diff
@@ -14,6 +14,31 @@
...
+/**
+ * Force plugins defined in a constant (supplied by phpunit.xml) to be active at runtime.
+ *
+ * @filter site_option_active_sitewide_plugins
+ * @filter option_active_plugins
+ *
+ * @param array $active_plugins
+ * @return array
+ */
+function xwp_filter_active_plugins_for_phpunit( $active_plugins ) {
```
kopepasah, line 26: @kidunot89 what was the reason to name this function `xwp_filter_...` instead of `stream_filter_...`?

tests/tests/connectors/test-class-connector-gravityforms.php
```diff
@@ -0,0 +1,949 @@
...
+	}
+
+	public function test_gravityforms_installed_and_activated() {
+		$this->assertTrue( class_exists( 'GFForms' ) );
+	}
+
+	public function create_form( $form = array(), $fields = array() ) {
+		$form    = array_merge(
+			$form,
+            array(
```
kopepasah, line 30: @kidunot89 this array tabbing looks off.

tests/bootstrap.php
kidunot89: @kopepasah It was copied from another plugin.

tests/tests/connectors/test-class-connector-gravityforms.php
```diff
@@ -0,0 +1,949 @@
...
+		);
+		$form_id = \GFAPI::add_form( $form );
+
+		return $form_id;
+	}
+
+	public function create_fields( $fields = array() ) {
+		$fields = ! empty( $fields ) ? $fields : array(
+			array(
+				'type' => 'text',
```
kopepasah, line 44: Let's get these assignments aligned.

tests/tests/connectors/test-class-connector-gravityforms.php
```diff
@@ -0,0 +1,949 @@
...
+			'name'        => 'Test confirmation',
+			'isDefault'   => true,
+			'type'        => 'message',
+			'message'     => 'Lorem ipsum dolor...',
+			'url'         => '',
+			'pageId'      => '',
+			'queryString' => '',
+		);
+
+		$form_id      = $this->create_form(
```
kopepasah, line 144: Align spacing.

tests/tests/connectors/test-class-connector-gravityforms.php
```diff
@@ -0,0 +1,949 @@
...
+					)
+				),
+				$this->greaterThan( 0 ),
+				$this->equalTo( 'forms' ),
+				$this->equalTo( 'updated' )
+			);
+
+		// Do stuff.
+		$notification = array(
+			'id'      => uniqid(),
```
kopepasah, line 173: Align assignments.

tests/tests/connectors/test-class-connector-gravityforms.php
```diff
@@ -0,0 +1,949 @@
...
+			'to'      => '{admin_email}',
+            'name'    => 'Test notification',
```
kopepasah, line 212: Align assignments.

tests/tests/connectors/test-class-connector-gravityforms.php
kopepasah: @kidunot89 realizing this is aligned with the $confirmation variable. Looks a bit weird. I will allow you to decide if it should stay.

tests/tests/connectors/test-class-connector-gravityforms.php
```diff
@@ -0,0 +1,949 @@
...
+				'payment_method'   => null,
+				'is_fulfilled'     => null,
+				'created_by'       => '1',
+				'transaction_type' => null,
+				'user_agent'       => 'Mozilla\/5.0 (Macintosh; Intel Mac OS X 10_11_1) AppleWebKit\/537.36 (KHTML, like Gecko) Chrome\/48.0.2564.116 Safari\/537.36',
+				'status'           => 'active',
+			)
+		);
+
+		// Wow, I actually had to do this just to get the Note ID.
```
kopepasah, line 475: @kidunot89 can you explain more about this comment and complexity? Is this something we need to be aware?

kopepasah (changes requested):

tests/tests/connectors/test-class-connector-gravityforms.php
kidunot89: Because the function doesn't return the ID and retrieving the it from `$wpdb` fails.

So I have to use a nearby hook in the function's logic and pass the ID back by reference.

composer.json
```diff
@@ -15,7 +15,8 @@
     "wp-cli/wp-cli-bundle": "^2.2",
     "wp-coding-standards/wpcs": "^2.2",
     "wp-phpunit/wp-phpunit": "^5.4",
-    "wpsh/local": "^0.2.3"
+    "wpsh/local": "^0.2.3",
+    "wp-premium/gravityforms": "^2.4.17"
```
kopepasah, line 19: @kidunot89 we need to find another way to pull in this plugin and test. While GPL code on in `wp-premium` Org is technically legal (I'm not a lawyer 🤐), we will not support it as it takes away from the developers who built these plugins and poses as a potential security risk.

kopepasah: Maybe we can pull it in as a zip someway? I have a developer's license, so access is not a problem.

kidunot89: I'm open to ideas, it's there a recommend method you have for pull this plugin in?

tests/tests/connectors/test-class-connector-gravityforms.php
kopepasah: Makes sense! 😀

Your comment had me curious and thanks for explaining!

composer.json
kopepasah: I am also open to ideas. One that comes to mind is we have a private S3bucket (or other storage service) from which we can read with private keys in Github. The we can use an archive installer with Composer to pull in those packages. 

I checked, but did not find any way to add files to Github or Travis for pulling in the data, but the above solution may be a good start.

@kasparsd any ideas?

tests/tests/connectors/test-class-connector-gravityforms.php
kidunot89: @kopepasah Later versions of the GravityForms have alternative to this that returns the Note ID.

composer.json
kasparsd: Honestly, I'm not sure how to handle this. How about we skip the tests if the dependency is not found for now?

kasparsd (commented):
@kidunot89 The tests alone look great. Could we get those merged in now and leave the plugin dependency out of tracking/dependency list until we find a solid solution for working with commercial plugins that don't have source code available publicly.

kidunot89:
@kasparsd I'm not sure about leaving the plugin out, since the tests relies on it.

derekherman:
@kidunot89 and @kasparsd In order to add the plugin it would need to be done in a way that is progressive and legal. It should not fail a build if the plugin is missing, so tests should only run if installed and active, and while we can install it with Composer I'm not sure if we can gracefully fail.

At any rate the minimum changes are that we need to add this object to the `repositories` array.

```
{
  "type": "package",
  "package": {
    "name": "gravityforms/gravityforms",
    "version": "2.4.21.3",
    "type": "wordpress-plugin",
    "dist": {
      "type": "zip",
      "url": "https://www.gravityhelp.com/wp-content/plugins/gravitymanager/api.php?op=get_plugin&slug=gravityforms&key={%GRAVITYFORMS_KEY}"
    },
    "require": {
      "composer/installers": "^1.4",
      "gotoandplay/gravityforms-composer-installer": "^2.3"
    }
  }
}
```

Also, we will need to add `"gravityforms/gravityforms": "2.4.21.3"` to `require-dev` and find a way to fail gracefully if the environment variable for the key it either missing or invalid. Locally you would also need to create a `.env` file and add `GRAVITYFORMS_KEY=YOUR_SECRET_KEY` and for the build to pass we can add the elite key we use for XWP projects to Travis CI.

There are really only two hurdles:
1. Make tests run with/without plugin
2. Degrade gracefully in Composer

kidunot89:
@kasparsd @ivankruchkoff The GravityForms package has been updated as @derekherman suggested, but I've added the `GRAVITYFORMS_KEY` as a secret in Travis-CI. If either of your reviews approve this should finally be all ready for merging.

composer.json
```diff
@@ -7,7 +7,23 @@
...
+    },
+    {
+      "type": "package",
+      "package": {
+        "name": "gravityforms/gravityforms",
+        "version": "2.4.21.3",
+        "type": "wordpress-plugin",
+        "dist": {
+          "type": "zip",
+          "url": "https://www.gravityhelp.com/wp-content/plugins/gravitymanager/api.php?op=get_plugin&slug=gravityforms&key={%GRAVITYFORMS_KEY}"
```
kasparsd, line 20: Does that mean that only people with GF key will be able to work on the plugin?

kasparsd (commented):
@kidunot89 I tried looking into the Travis issue but I'm not sure what is happening. Could it be that some of the dependencies were updated via `composer update`? It appears to be happening here:

https://github.com/symfony/polyfill-php80/blob/dc3063ba22c2a1fd2f45ed856374d79114998f91/bootstrap.php#L23

composer.json
kidunot89: @kasparsd Yea, it is. :thinking: This might need some further tooling.

kidunot89: I think we should skip GravityForms unless the develop installs it as a `dev` specifically with composer. We can probably move the whole GravityForms install process into the Travis-CI script.

kidunot89: @kasparsd :point_up: What do you think?

kasparsd: Yes, that is a good idea! We could even have another composer.json file for installing these external dependencies.

.travis.yml
```diff
@@ -71,7 +72,25 @@ before_install:
...
+          \"url\": \"https://www.gravityhelp.com/wp-content/plugins/gravitymanager/api.php?op=get_plugin&slug=gravityforms&key=$GRAVITYFORMS_KEY\"
+        },
+        \"require\": {
+          \"composer/installers\": \"^1.4\",
+          \"gotoandplay/gravityforms-composer-installer\": \"^2.3\"
+        }
+      }
+    }"
   - npm install
+  - composer require --dev gravityforms/gravityforms:2.4.21.3
```
kidunot89, line 93: @kasparsd This should do the trick. I don't think there is a need for an alternative `composer.json`, but it does make the repositories array and object.
image
But I revert it in the `before_deploy` :thinking:.

.travis.yml
```diff
@@ -50,6 +50,7 @@ jobs:
       if: branch IS present AND type = push
       script: npm run release
       before_deploy:
+        - git restore composer.json
```
kidunot89, line 53: @kasparsd This is to revert the changes made in the `install` below.

PR #1149, opened 2020-07-28
Record timestamp creation refactored

### LINKED ISSUE #1140: Stream Records dates are 2052 years ago with zeroed 'created' fields in database
# Bug Report

The dates all display as:

```
2052 years ago
-0001/11/30
12:00:00 AM
```

And also, in the database, the created field in wp_stream are all zeroes.

## Screenshots

IMG_3659

IMG_3660

Drag and drop screenshots in this area that help illustrate the problem.

## System Information

- Stream plugin version: 3.5.0
- WordPress version: 5.4.2
- PHP version: 7.4
- Browser: iPadOS Safari
- I also use MySQL 8 on DigitalOcean's managed database service

kidunot89:
Fixes #1140 

# Summary
This replaces the usage of `wp_stream_get_iso_8601_extended_date()` for creating record timestamp with `current_time()`

### DIFF
--- a/classes/class-log.php
+++ b/classes/class-log.php
@@ -120,9 +120,6 @@ class Log {
 		// Add user meta to Stream meta.
 		$stream_meta['user_meta'] = $user_meta;
 
-		// Get the current time in milliseconds.
-		$iso_8601_extended_date = wp_stream_get_iso_8601_extended_date();
-
 		if ( ! empty( $user->roles ) ) {
 			$roles = array_values( $user->roles );
 			$role  = $roles[0];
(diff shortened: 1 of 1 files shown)


classes/class-log.php
```diff
@@ -138,7 +135,7 @@ function ( $var ) {
 			'blog_id'   => (int) apply_filters( 'wp_stream_blog_id_logged', get_current_blog_id() ),
 			'user_id'   => (int) $user_id,
 			'user_role' => (string) $role,
-			'created'   => (string) $iso_8601_extended_date,
+			'created'   => (string) current_time( 'mysql', true ),
```
kasparsd, line 138: We always need the UTC timestamp formatted for the [`DATETIME` field](https://dev.mysql.com/doc/refman/8.0/en/datetime.html) in the DB so this will ensure we always get that consistent value.

This is different from the current behaviour of `wp_stream_get_iso_8601_extended_date()` which always adds the timezone information:

https://github.com/xwp/stream/blob/4f85afb87e007c6c968859b821a399f4c9ae2b42/includes/functions.php#L62

which was probably causing the bug as the MySQL client wasn't able to cast it to a valid `DATETIME` value.

PR #1151, opened 2020-07-30
Users connector class test implemented.

### LINKED ISSUE #946: Stream not capturing SSO sign-ins from OneLogin
We use the [OneLogin plugin](https://wordpress.org/plugins/onelogin-saml-sso/) for authentication on our site, and Stream is not capturing user log in events. These events are recorded perfectly on clones of the site that use the standard WordPress authentication.

Happy to help dig up logs or other necessary data for testing.

kidunot89:
Partially fixes #1093 
Potentially fixes #946 

# Summary checklist
- [x] `clear_auth_cookie` callback test implemented and passing.
- [x] `delete_user` callback test implemented and passing.
- [x] `deleted_user` callback test implemented and passing.
- [x] `password_reset` callback test implemented and passing.
- [x] `profile_update` callback test implemented and passing.
- [x] `retrieve_password` callback test implemented and passing.
- [x] `set_current_user` callback test implemented and passing.
- [x] `set_logged_in_cookie` callback test implemented and passing.
- [x] `set_user_role` callback test implemented and passing.
- [x] `user_register` callback test implemented and passing.

### DIFF
--- a/classes/class-connector.php
+++ b/classes/class-connector.php
@@ -91,10 +91,6 @@ abstract class Connector {
 	 * Unregister all context hooks
 	 */
 	public function unregister() {
-		if ( ! $this->is_registered ) {
-			return;
-		}
-
 		foreach ( $this->actions as $action ) {
 			remove_action( $action, array( $this, 'callback' ), 10, 99 );
 		}
--- a/classes/class-connector.php
+++ b/classes/class-connector.php
@@ -91,6 +91,10 @@ abstract class Connector {
 	 * Unregister all context hooks
 	 */
 	public function unregister() {
+		if ( ! $this->is_registered ) {
+			return;
+		}
+
 		foreach ( $this->actions as $action ) {
 			remove_action( $action, array( $this, 'callback' ), 10, 99 );
 		}
--- a/connectors/class-connector-users.php
+++ b/connectors/class-connector-users.php
@@ -41,6 +41,7 @@ class Connector_Users extends Connector {
 		'delete_user',
 		'deleted_user',
 		'set_user_role',
+		'set_current_user',
 	);
 
 	/**
@@ -242,7 +243,7 @@ class Connector_Users extends Connector {
 			/* translators: %s: a user display name (e.g. "Jane Doe") */
 			__( '%s\'s password was reset', 'stream' ),
 			array(
-				'email' => $user->display_name,
+				'display_name' => $user->display_name,
 			),
 			$user->ID,
 			'profiles',
@@ -384,6 +385,31 @@ class Connector_Users extends Connector {
 			$user_id,
 			'users',
 			'deleted',
+			$user->ID
+		);
+	}
+
+	/**
+	 * Logs password-less user login
+	 *
+	 * @action set_current_user
+	 */
+	public function callback_set_current_user() {
+		$user_id = get_current_user_id();
+
+		if ( 0 === $user_id ) {
+			return;
+		}
+
+		$user = get_user_by( 'id', $user_id );
+
+		$this->log(
+			/* translators: %s: a user display name (e.g. "Jane Doe") */
+			__( '%s logged in', 'stream' ),
+			array( 'display_name' => $user->display_name ),
+			$user->ID,
+			'sessions',
+			'login',
 			$user->ID
 		);
 	}
--- a/connectors/class-connector-users.php
+++ b/connectors/class-connector-users.php
@@ -317,7 +317,6 @@ class Connector_Users extends Connector {
 		$user_id = get_current_user_id();
 		$user    = get_user_by( 'ID', $user_id );
 
-		error_log( print_r( $user, true ) );
 		// For some reason, incognito mode calls clear_auth_cookie on failed login attempts.
 		if ( empty( $user ) || ! $user->exists() ) {
 			return;
--- a/connectors/class-connector-users.php
+++ b/connectors/class-connector-users.php
@@ -314,8 +314,10 @@ class Connector_Users extends Connector {
 	 * @action clear_auth_cookie
 	 */
 	public function callback_clear_auth_cookie() {
-		$user = wp_get_current_user();
+		$user_id = get_current_user_id();
+		$user    = get_user_by( 'ID', $user_id );
 
+		error_log( print_r( $user, true ) );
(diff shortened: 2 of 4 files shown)


connectors/class-connector-users.php
```diff
@@ -387,4 +388,29 @@ public function callback_deleted_user( $user_id ) {
...
+
+		if ( 0 === $user_id ) {
+			return;
+		}
+
+		$user = get_user_by( 'id', $user_id );
+
+		$this->log(
+			/* translators: %s: a user display name (e.g. "Jane Doe") */
+			__( '%s logged in', 'stream' ),
```
dero, line 408: @kidunot89 I would recommend changing the wording here, because `wp_set_current_user` can be used to achieve a variety of effects, not just to facilitate authentication via external services.

Maybe something like: `'%s set as the active user'`?

tests/tests/connectors/test-class-connector-users.php
```diff
@@ -0,0 +1,307 @@
...
+
+		// Make partial of Connector_Users class, with mocked "log" function.
+		$this->mock = $this->getMockBuilder( Connector_Users::class )
+			->setMethods( array( 'log' ) )
+			->getMock();
+
+		$this->mock->register();
+	}
+
+	public function test_callback_user_register() {
```
dero, line 20: This is a good testing flow. 👍

connectors/class-connector-users.php
dero: Also, should we also respond to `wp_set_current_user( 0 )` – something like `'Active user unset'`?

tests/tests/connectors/test-class-connector-users.php
```diff
@@ -0,0 +1,307 @@
...
+					$this->equalTo( $user_id ),
+					$this->equalTo( 'sessions' ),
+					$this->equalTo( 'login' ),
+					$this->equalTo( $user_id )
+				)
+			);
+
+		// Do stuff.
+		wp_set_current_user( $user_id );
+		wp_set_current_user( 0 );
```
dero, line 301: @kidunot89 If the previous call generates an entry in the log, then ideally calling the same function with `0` as an argument generates an entry as well, otherwise we'll end up with two subsequent "logged in" events and no "log out". See previous comments, please.

dero (changes requested):
@kidunot89 This looks really good, just a couple comments here. Thanks!

classes/class-connector.php
```diff
@@ -91,10 +91,6 @@ public function register() {
 	 * Unregister all context hooks
 	 */
 	public function unregister() {
-		if ( ! $this->is_registered ) {
```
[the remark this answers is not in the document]
kidunot89: @dero Because I noticed that somewhere between tests the connectors where ending in a state where the hooks were `is_registered` flag was `false` but the action were still registered and somehow prevent the mocked connector class actions from getting the data on logs catching deletions. This has actually been happening across multiple connector test and is also the cause for hold up on the last test of the comment connector test case

connectors/class-connector-users.php
kidunot89: I don't think so because `wp_set_current_user( 0 )` is used to end the user session typically.

tests/tests/connectors/test-class-connector-users.php
kidunot89: Not necessary, Stream is designed to target adminstrative actions primarily

connectors/class-connector-users.php
dero: @kidunot89 OK, fair enough.

What do you think about the other suggestion? If Stream is designed to capture administrative actions, should we even observe `set_current_user`, which is generally used programatically?

kidunot89: Some plugins like OneLogin that rely `wp_set_current_user()` to authenticate to user. See this issue [here](https://github.com/xwp/stream/issues/946)

dero: > See this issue here.

I'm aware of the issue and the _intent_ here, @kidunot89 – what I'm saying is `wp_set_current_user` is used in many contexts and it is being used programmatically. So:

* If Stream is supposed to mainly track administrative actions, observing `set_current_users` will lead to tracking many programmatic actions instead and to reporting log ins where no log ins are actually happening.
* If we want to track these programmatic actions as well, we should track its opposite as well, i.e. `wp_set_current_user( 0 )`.

As for SSO plugins, we should probably deal with them on individual basis. OneLogin, for instance, uses this code:

```php
if (is_a($user_id, 'WP_Error')) {
	$errors = $user_id->get_error_messages();
	foreach($errors as $error) {
		echo esc_html($error).'<br>';
	}
	exit();
} else if ($user_id) {
	wp_set_current_user($user_id);
	
	$rememberme = false;
	$remembermeMapping = get_option('onelogin_saml_attr_mapping_rememberme');
	if (!empty($remembermeMapping) && isset($attrs[$remembermeMapping]) && !empty($attrs[$remembermeMapping][0])) {
		$rememberme = in_array($attrs[$remembermeMapping][0], array(1, true, '1', 'yes', 'on')) ? true : false;
	}
	wp_set_auth_cookie($user_id, $rememberme);

	setcookie(SAML_LOGIN_COOKIE, 1, time() + YEAR_IN_SECONDS, SITECOOKIEPATH );
	setcookie(SAML_NAMEID_COOKIE, $auth->getNameId(), time() + YEAR_IN_SECONDS, SITECOOKIEPATH );
	setcookie(SAML_SESSIONINDEX_COOKIE, $auth->getSessionIndex(), time() + YEAR_IN_SECONDS, SITECOOKIEPATH );
	setcookie(SAML_NAMEID_FORMAT_COOKIE, $auth->getNameIdFormat(), time() + YEAR_IN_SECONDS, SITECOOKIEPATH );
}
```

... which means that whenever `wp_set_current_user` is called, `wp_set_auth_cookie` is called as well, which in turn runs the `set_logged_in_cookie` action. There is apparently no reason why Stream would not track those log-ins.

Have you investigated why the cookie action is not run? Seeing that the issue is rather old, is it possible the plugin code changed in the meantime?

kidunot89: Possibly 🤔

kidunot89: @dero I believe you are right. This callback would add too much noise and unpredictability to the records.

PR #1154, opened 2020-08-05
Media connector test implemented.

### LINKED ISSUE #907: Media connector assumes parent post exists
In `Connector_Media`, [the code](https://github.com/xwp/stream/blob/c284f36acc094dd579720f3130e646e097aba9b6/connectors/class-connector-media.php#L140-L141) assumes that if there's a parent ID, the parent post exists. This is not always true if you have an orphaned attachment.

I'd suggest changing the ternary to use `$parent` instead of `$parent_id`

kidunot89:
Partially fixes #1093 
Fixes #907 

# Summary Checklist
- [x] `add_attachment` callback test implemented and passing. Refactored to avoid breaking on invalid parent posts.
- [x] `delete_attachment` callback test implemented and passing.
- [x] `edit_attachment` callback test implemented and passing.
- [x] `wp_save_image_editor_file` callback test implemented and passing.
- No test created for the `wp_save_image_file` callback because the hook is deprecated in WordPress and `wp_save_image_editor_file` is the replacement.

#907 patch implemented here to avoid future merge conflicts if put in separate PR.

### DIFF
--- a/connectors/class-connector-media.php
+++ b/connectors/class-connector-media.php
@@ -151,7 +151,7 @@ class Connector_Media extends Connector {
 		$url             = $post->guid;
 		$parent_id       = $post->post_parent;
 		$parent          = get_post( $parent_id );
-		$parent_title    = $parent_id ? $parent->post_title : null;
+		$parent_title    = $parent instanceof \WP_Post ? $parent->post_title : 'Unidentifiable post';
 		$attachment_type = $this->get_attachment_type( $post->guid );
 
 		$this->log(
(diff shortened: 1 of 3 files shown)


connectors/class-connector-media.php
```diff
@@ -151,7 +151,7 @@ public function callback_add_attachment( $post_id ) {
 		$url             = $post->guid;
 		$parent_id       = $post->post_parent;
 		$parent          = get_post( $parent_id );
-		$parent_title    = $parent_id ? $parent->post_title : null;
+		$parent_title    = $parent instanceof \WP_Post ? $parent->post_title : 'Unidentifiable post';
```
kidunot89, line 154: @kasparsd Please take a look at this :point_up:. It resolves #907. I also updated the corresponding test to test this logic as well.

PR #1155, opened 2020-08-12
Installer connector test implemented

kidunot89:
Partially fixes #1093 

# Summary Checklist
- [ ] ~`upgrader_process_complete` callback test implemented and passing.~ **Skipped**
- [x] `activate_plugin` callback test implemented and passing. 
- [x] `deactivate_plugin` callback test implemented and passing.
- [x] `delete_site_transient_update_themes` callback test implemented and passing.
- [ ] ~`pre_set_site_transient_update_plugins` callback test implemented and passing.~ **Skipped**
- [ ] ~`_core_updated_successfully` callback test implemented and passing.~ **Skipped**
- [x] `switch_theme` callback test implemented and passing.

Some tests are incomplete and skipped because I can't find a way to simulate the scenarios needed.

### DIFF
--- a/tests/tests/connectors/test-class-connector-installer.php
+++ b/tests/tests/connectors/test-class-connector-installer.php
@@ -23,15 +23,53 @@ class Test_WP_Stream_Connector_Installer extends WP_StreamTestCase {
 
 	public function test_callback_upgrader_process_complete() {
 		// Prepare scenario
+		$this->markTestSkipped( 'This test skipped until the needed scenario can be properly simulated.' );
 
 		// Expected log calls.
 		$this->mock->expects( $this->once() )
 			->method( 'log' )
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
     image: mysql:5
     volumes:
       - db_data:/var/lib/mysql
-    restart: always
+    restart: unless-stopped
```
kasparsd, line 9: This appears to not start the container if it [has been stopped manually](https://docs.docker.com/compose/compose-file/#restart). Are we sure this is not going to cause unexpected results during test runs, for example?

kidunot89: I'll remove it, just wanted an alternative to `always`.

PR #1159, opened 2020-08-20
Default options provided to "wp_stream_auto_purge" callbacks.

### LINKED ISSUE #1056: Old records are not being removed

kidunot89:
Fixes #1056

# Checklist
Provides `Settings::default_settings()` as the default value for all `get_site_option( 'wp_stream_network' );` and `get_option( 'wp_stream' );` calls to avoid skipping record purging when no settings are saved in the database, which seems to be the root cause of #1056.


laxbobber:
Any idea when this will be included in a release? I've got 150 sites piling up records that I would love to purge.

kidunot89:
@kasparsd @dero :point_up:

PR #1160, opened 2020-08-20
WHERE statement filter added.

### LINKED ISSUE #1044: Add a filter to where db clause
Stream allows us to modify the final query via `wp_stream_db_query`. It would be helpful to have separate filters for `$where` and `$join` clauses. Otherwise its hard to coordinate multiple where and join queries added . by separate plugins.

A use case for this is multiple plugins can add multiple filters to stream (based on meta value etc). A plugin may not be aware the way query was modified so they can add their own conditions

```
		/**
		 * BUILD THE FINAL QUERY
		 */
		$query = "SELECT SQL_CALC_FOUND_ROWS {$select}
		FROM $wpdb->stream
		{$join}
		WHERE 1=1 {$where}
		{$orderby}
		{$limits}";

		/**
		 * Filter allows the final query to be modified before execution
		 *
		 * @param string $query
		 * @param array  $args
		 *
		 * @return string
		 */
		$query = apply_filters( 'wp_stream_db_query', $query, $args );
```

A quick solution would be

```
		/**
		 * BUILD THE FINAL QUERY
		 */

                // We can directly pass the where query or get all the where condition in array and join them
                $where = apply_filters( `wp_stream_db_query`, $where );


		$query = "SELECT SQL_CALC_FOUND_ROWS {$select}
		FROM $wpdb->stream
		{$join}
		WHERE 1=1 {$where}
		{$orderby}
		{$limits}";

		/**
		 * Filter allows the final query to be modified before execution
		 *
		 * @param string $query
		 * @param array  $args
		 *
		 * @return string
		 */
		$query = apply_filters( 'wp_stream_db_query', $query, $args );
```

kidunot89:
Fixes #1044 .

# Checklist
Adds a filter for modifying the WHERE statement and the records query.

### DIFF
--- a/classes/class-query.php
+++ b/classes/class-query.php
@@ -217,6 +217,16 @@ class Query {
 		$select = implode( ', ', $selects );
 
 		/**
+		 * Filters query WHERE statement as an alternative to filtering
+		 * the $query using the hook below.
+		 *
+		 * @param string $where  WHERE statement.
+		 *
+		 * @return string
+		 */
+		$where = apply_filters( 'wp_stream_db_query', $where );
+
+		/**
 		 * BUILD THE FINAL QUERY
 		 */
 		$query = "SELECT SQL_CALC_FOUND_ROWS {$select}
--- a/classes/class-query.php
+++ b/classes/class-query.php
@@ -224,7 +224,7 @@ class Query {
 		 *
 		 * @return string
 		 */
-		$where = apply_filters( 'wp_stream_db_query', $where );
+		$where = apply_filters( 'wp_stream_db_query_where', $where );
 
 		/**
 		 * BUILD THE FINAL QUERY


classes/class-query.php
```diff
@@ -216,6 +216,16 @@ public function query( $args ) {
...
 
+		/**
+		 * Filters query WHERE statement as an alternative to filtering
+		 * the $query using the hook below.
+		 *
+		 * @param string $where  WHERE statement.
+		 *
+		 * @return string
+		 */
+		$where = apply_filters( 'wp_stream_db_query', $where );
```
dero, line 227: @kidunot89 I would expect the filter name to be different from `wp_stream_db_query`, which is already used below. How about `wp_stream_db_query_where` instead?

dero (changes requested):
@kidunot89 Thanks. One question / suggestion.

classes/class-query.php
kidunot89: :man_facepalming: my bad

kidunot89:
@dero you're requested changes have been made. Please re-review.

dero (approved):
LGTM now. Thanks!

PR #1163, opened 2020-08-27
Automatically tag releases in the distribution repository

### LINKED ISSUE #1157: Release tagging fails on tags
The release script fails to commit a tagged release because it checks out a branch with the exact same name during the process. This causes the `git push` to find two references of the same name:

> error: src refspec 3.5.1 matches more than one

from https://travis-ci.com/github/xwp/stream/jobs/372078948#L474

kidunot89:
Fixes #1157.

# Summary
Add `git push --delete origin "$DIST_TAG"` statement after new tag creation in the distribution script to bypass the tag naming conflict during final `git push ...`call.  

@kasparsd Not sure how to test this. :man_shrugging:

### DIFF
--- a/local/scripts/dist.sh
+++ b/local/scripts/dist.sh
@@ -33,6 +33,7 @@ git commit --allow-empty --message "$COMMIT_MESSAGE"
 if [ -n "$DIST_TAG" ]; then
 	echo "Tagging a release: $DIST_TAG"
 	git tag --force "$DIST_TAG"
+	git push --delete origin "$DIST_TAG"
 fi
 
 git push --force --set-upstream origin "$DIST_BRANCH" --tags
--- a/local/scripts/dist.sh
+++ b/local/scripts/dist.sh
@@ -33,7 +33,8 @@ git commit --allow-empty --message "$COMMIT_MESSAGE"
 if [ -n "$DIST_TAG" ]; then
 	echo "Tagging a release: $DIST_TAG"
 	git tag --force "$DIST_TAG"
-	git push --delete origin "$DIST_TAG"
+	git push --force --set-upstream origin --tags
+else
+	# Syncing release artifacts for a branch.
+	git push --force --set-upstream origin "$DIST_BRANCH"
 fi
-
-git push --force --set-upstream origin "$DIST_BRANCH" --tags
--- a/classes/class-list-table.php
+++ b/classes/class-list-table.php
@@ -99,7 +99,7 @@ class List_Table extends \WP_List_Table {
 	public function no_items() {
 		?>
 		<div class="stream-list-table-no-items">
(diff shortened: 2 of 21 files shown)


local/scripts/dist.sh
```diff
@@ -33,6 +33,7 @@ git commit --allow-empty --message "$COMMIT_MESSAGE"
 if [ -n "$DIST_TAG" ]; then
 	echo "Tagging a release: $DIST_TAG"
 	git tag --force "$DIST_TAG"
+	git push --delete origin "$DIST_TAG"
```
dero, line 36: @kidunot89 AFAICT this will delete the current branch remotely, but will keep it locally along with the tag of the same name. Effectively leading to the same issue as before (?), where the `$DIST_BRANCH` refers both to a local tag and a local branch name.

Can you or @kasparsd confirm what the requirement is here?

The simplest solution would be name tags and branches differently, e.g. name the tag simply `3.5.1` and the branch `version-3.5.1`. Whether you want to keep it after a release or not is not really relevant. Having two names not only helps to disambiguate the ref types, but could also potentially help with debugging issues.

If we don't want to disambiguate the two, we could perhaps do this:

```
git push --force --set-upstream origin "refs/heads/${DIST_BRANCH}:refs/heads/{DIST_BRANCH}" --tags
```

dero (commented):
Adding a question here, @kidunot89.

local/scripts/dist.sh
kidunot89: @dero I believe the goal is to rebuild the distribution tag with all needed assets for packagist.org

kidunot89: This [error](https://travis-ci.com/github/xwp/stream/jobs/372078948#L474) was being caused by a collision with the remote tag when the rebuilt tag is pushed to the repo on the final line of the script.

kidunot89: @dero See [this](https://github.com/xwp/stream/pull/1074)

local/scripts/dist.sh
```diff
@@ -33,6 +33,8 @@ git commit --allow-empty --message "$COMMIT_MESSAGE"
 if [ -n "$DIST_TAG" ]; then
 	echo "Tagging a release: $DIST_TAG"
 	git tag --force "$DIST_TAG"
+	git push --force --set-upstream origin --tags
```
kasparsd, line 36: @dero I think the issue was caused by the `$DIST_BRANCH` being used together with the `--tags` flag. Now it will force-push all tags or just the specific branch.

I've tested this locally and it appears to be working as expected. Since this particular file isn't included in the release bundle I suggest we merge this in and test during the next release.

dero: Yes, looks good now. :+1:

dero (approved):

PR #1165, opened 2020-08-28
Settings connector test implemented.

kidunot89:
Fixes #1093 .

Summary Checklist
- Test added for all connector callbacks.
- `callback_whitelist_options` replaced with `callback_allowed_options` due to the `whitelist_options` callback being deprecated for the `allowed_options` filter.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,8 +6,8 @@
   "license": "GPL-2.0-or-later",
   "repositories": [
     {
-      "type":"composer",
-      "url":"https://wpackagist.org"
+      "type": "composer",
+      "url": "https://wpackagist.org"
     }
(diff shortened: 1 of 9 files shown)


dero (commented):
@kidunot89 This looks great! One request: can you please update `callback_update_option` to take parameters in a correct order?

Right now the method signature is:

```php
public function callback_update_option( $option, $value, $old_value )
```

... but the `update_option` hook [passes the `$old_value` first](https://developer.wordpress.org/reference/hooks/update_option/). I believe the only thing we need to do is to switch the names of the variables there.

kidunot89:
Fixed :+1:

PR #1177, opened 2020-09-23
Blog connector test implemented and connector updated.

### LINKED ISSUE #1175: Site deletion isn't logged
# Bug Report

## Expected Behavior

When a site gets deleted from a Multisite network, the actual site deletion isn't logged in Stream. Several related records are logged instead.

## Actual Behavior

When a site gets deleted, Stream logs a record for every user who gets removed from the site, and logs a record that the "Blog Count" setting was updated, but there's no record of the site being deleted.

1. From the Network Admin -> Sites screen, click the `Delete` action link for a site
2. Confirm your action
3. View the Stream logs for the network

I think the preference here would be for a single "Site Deleted" log to be recorded in place of all the logs that are recorded for each user on the site being removed from it. If a site with 100 users gets deleted, you have 101 records creating unnecessary noise.

### LINKED ISSUE #1187: siteurl on integration data for Site create/update/delete
# Feature Request

On Sites log and data should have the siteurl. My use case, for example, I use that to dinamically create an register on Cloudflare with the subdomain. But the subdomain are not sent on the integration with my webhook (I used Slack option as webhook).

For solve my problem now I created a fork replacing sitename to siteurl, that's it's more usefull to me.

## System Information

- Stream plugin version: 3.5.1
- WordPress version: 5.5.1
- PHP version: 7.4
- Browser: Chrome
- Computer operating system: Windows

kidunot89:
Works towards #1093 .
Fixes #1175 
Fixes #1187  

# Summary checklist
- [x] `wpmu_new_blog` callback replaced with `wp_initialize_site` callback due to the deprecation of the `wpmu_new_blog` hook.
- [x] `wp_initialize_site` callback tested and passing.
- [x] `wp_delete_site` callback implemented, testing and passing.
- [x] `wpmu_activate_blog` callback tested and passing.
- [x] `add_user_to_blog` callback tested and passing.
- [x] `remove_user_from_blog` callback tested and passing.
- [x] `update_blog status` function and all connected hooks tested and passing.

@dero @kasparsd I updated the Blog connector log meta with more info. It know include the the blog `name`, `url`, and `id`.

### DIFF
--- a/connectors/class-connector-blogs.php
+++ b/connectors/class-connector-blogs.php
@@ -363,7 +363,6 @@ class Connector_Blogs extends Connector {
 	 */
 	public function callback_update_blog_status( $blog_id, $status, $action ) {
 		$blog = get_site( $blog_id );
-
 		$this->log(
 			/* translators: %1$s: a site name, %2$s: a blog status (e.g. "FooBar Blog", "archived") */
 			_x(
--- a/connectors/class-connector-blogs.php
+++ b/connectors/class-connector-blogs.php
@@ -24,7 +24,8 @@ class Connector_Blogs extends Connector {
 	 * @var array
 	 */
 	public $actions = array(
-		'wp_insert_site',
+		'wp_initialize_site',
+		'wp_delete_site',
 		'wpmu_activate_blog',
 		'wpmu_new_user',
 		'add_user_to_blog',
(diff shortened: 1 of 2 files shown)


connectors/class-connector-blogs.php
```diff
@@ -144,14 +147,38 @@ public function callback_wpmu_new_blog( $blog_id ) {
...
+			),
+			array(
+				'site_name' => $old_site->blogname,
+			),
+			$old_site->blog_id,
+			sanitize_key( $old_site->blogname ),
+			'deleted'
+		);
+	}
+
```
kidunot89, line 181: @dero @johnbillion This callback was added to resolve #1175. Please take a look.

dero: Thanks @kidunot89 -- this callback will add an appropriate log entry, but AFAICT won't prevent the related "User removed" etc. entries from being logged, which can really clutter the log when a blog with dozens of users is removed.

I'm OK approving this PR as is, but I think we should keep #1175 open until we're able to bundle up those excessive individual log entries somehow. I'm thinking perhaps we find a way to just log "XYZ users were removed from the ABC site."?

What do you guys think?

dero (commented):
@kidunot89 Good work here, thanks!

One question related to #1175

PR #1182, opened 2020-09-29
Record list table "No items" message updated

### LINKED ISSUE #1178: Stream shouldn't apologise when there are no records
# Bug Report

## Expected Behavior

When there are no records to show, the messaging to the user should be clear.

## Actual Behavior

Stream apologises to the user: `Sorry, no activity records were found`. It's a small issue but there is no reason for Stream to apologise here, and for a non-native English speaker this message may not be clear and may suggest there was a problem fetching the records, when in fact everything is working as expected.

I recommend that the `Sorry` prefix is removed from this phrase to enhance the clarity.

1. Clear your Stream database from the Stream -> Settings -> Advanced menu
2. View the activity log
3. See an apology

kidunot89:
Fixes #1178.

Changes "no_items" message for the Records list table to the message in the capture below.

image

### DIFF
--- a/classes/class-list-table.php
+++ b/classes/class-list-table.php
@@ -99,7 +99,7 @@ class List_Table extends \WP_List_Table {
 	public function no_items() {
 		?>
 		<div class="stream-list-table-no-items">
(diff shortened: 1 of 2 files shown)


dero (approved):
LGTM

PR #1185, opened 2020-10-01
Fix: column value getting overridden when multiple custom columns are added

### LINKED ISSUE #1184: Adding multiple columns to the stream table using filters only display the last column correctly
# Bug Report

All the columns before the last one just display the column title as the column value.

Here is a sample code which can be tried to reproduce this case - 

```PHP
add_filter(
	'wp_stream_list_table_columns',
	function ( array $columns ) : array {
		$columns['test_column_1'] = 'test_column_1';
		$columns['test_column_2'] = 'test_column_2';

		return $columns;
	}
);

add_filter(
	'wp_stream_insert_column_default_test_column_1',
	function ( string $column_name, \WP_Stream\Record $record ) : string {
		return 'test_column_1  - ' . $record->ID;
	},
	10,
	2
);

add_filter(
	'wp_stream_insert_column_default_test_column_2',
	function ( string $column_name, \WP_Stream\Record $record ) : string {
		return 'test_column_2 - ' . $record->ID;
	}
	,
	10,
	2
);

add_filter(
	'wp_stream_register_column_defaults',
	function ( array $new_columns ) : array {
		$new_columns[] = 'test_column_1';
		$new_columns[] = 'test_column_2';

		return $new_columns;
	}
);
```

In this snippet, I am adding two columns to the stream table but the first column only displays the column name `test_column_1` in all the columns and the second column displays the correct value `test_column_2 - <record-id>`

Any ideas on how this can be fixed?

## Expected Behavior

All columns that are added display the correct expected value.

## Actual Behavior

Only the last column displays the correct value for the record.

Nikschavan:
Fixes #1184.

The statement `$out = $column_name;` would get called even after the correct value is set to the variable for the first column. 

I have changed this so that the default value will be used only if the variable `out` is still unchanged so that it retains the correct column value for all the columns.

# Checklist

## Release Changelog

- Fix: Column value getting overridden when multiple custom columns are added.

## Release Checklist

### DIFF
--- a/classes/class-list-table.php
+++ b/classes/class-list-table.php
@@ -396,6 +396,12 @@ class List_Table extends \WP_List_Table {
 
 				if ( ! empty( $inserted_columns ) && is_array( $inserted_columns ) ) {
 					foreach ( $inserted_columns as $column_title ) {
+
+						// Set default value of column to column name.
+						if ( empty( $out ) ) {
+							$out = $column_name;
+						}
+
 						/**
 						 * If column title inserted via wp_stream_register_column_defaults ($column_title) exists
 						 * among columns registered with get_columns ($column_name) and there is an action associated
@@ -413,8 +419,6 @@ class List_Table extends \WP_List_Table {
 							 * @return string
 							 */
 							$out = apply_filters( "wp_stream_insert_column_default_{$column_title}", $column_name, $record );
-						} else {
-							$out = $column_name;
 						}
 					}
 				} else {
--- a/classes/class-list-table.php
+++ b/classes/class-list-table.php
@@ -389,19 +389,14 @@ class List_Table extends \WP_List_Table {
 				 * Registers new Columns to be inserted into the table. The cell contents of this column is set
 				 * below with 'wp_stream_insert_column_default_'
 				 *
+				 * @param array $new_columns  Columns injected in the table.
+				 *
 				 * @return array
 				 */
-				$new_columns      = array();
-				$inserted_columns = apply_filters( 'wp_stream_register_column_defaults', $new_columns );
+				$inserted_columns = apply_filters( 'wp_stream_register_column_defaults', array() );
 
 				if ( ! empty( $inserted_columns ) && is_array( $inserted_columns ) ) {
 					foreach ( $inserted_columns as $column_title ) {
-
-						// Set default value of column to column name.
-						if ( empty( $out ) ) {
-							$out = $column_name;
-						}
-
 						/**
 						 * If column title inserted via wp_stream_register_column_defaults ($column_title) exists
 						 * among columns registered with get_columns ($column_name) and there is an action associated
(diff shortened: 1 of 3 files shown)


Nikschavan:
Hi @kidunot89, @dero, @kasparsd 👋 

Any chance to get a review on this?

classes/class-list-table.php
```diff
@@ -396,6 +396,12 @@ public function column_default( $item, $column_name ) {
 
 				if ( ! empty( $inserted_columns ) && is_array( $inserted_columns ) ) {
 					foreach ( $inserted_columns as $column_title ) {
+
```
derekherman, line 399: @kasparsd or @kidunot89 Correct me if I'm wrong here, but to me this seems like the problem is that the loop needs to exit if there is a value found and also that the default gets set if there isn't a value coming from the filters. 

So the whole code block would change to:
```
if ( ! empty( $inserted_columns ) && is_array( $inserted_columns ) ) {
	foreach ( $inserted_columns as $column_title ) {

		/**
		 * If column title inserted via wp_stream_register_column_defaults ($column_title) exists
		 * among columns registered with get_columns ($column_name) and there is an action associated
		 * with this column, do the action
		 *
		 * Also, note that the action name must include the $column_title registered
		 * with wp_stream_register_column_defaults
		 */
		if ( $column_title === $column_name && has_filter( "wp_stream_insert_column_default_{$column_title}" ) ) {
			/**
			 * Allows for the addition of content under a specified column.
			 *
			 * @param object $record Contents of the row
			 *
			 * @return string
			 */
			$out = apply_filters( "wp_stream_insert_column_default_{$column_title}", $column_name, $record );
			break;
		}
	}
}

// Set default value.
if ( empty( $out ) ) {
	$out = $column_name;
}
```

classes/class-list-table.php
```diff
@@ -396,6 +396,12 @@ public function column_default( $item, $column_name ) {
...
+						// Set default value of column to column name.
+						if ( empty( $out ) ) {
+							$out = $column_name;
+						}
+
```
kidunot89, line 404: As @derekherman stated above the loop has to be exited once the value is found or it will just overwrite itself with the default value.

@Nikschavan You can replace the whole conditional started on Line 397 with @derekherman suggestion above and that should work.

kidunot89 (changes requested):
@Nikschavan I'm sorry it took so long to get this review :disappointed:.

classes/class-list-table.php
Nikschavan: This should work as expected, I will update the PR

One question - If any column value is actually `empty` (for an example case because that column is not relevant for that log type) that gets set to `column_name` everywhere, That is because of the last `// Set default value.` section. Is this an expected behaviour or can that be removed?

derekherman: I’m not sure what the intended default behavior is. Is there a test for this @kidunot89

kidunot89: @derekherman @Nikschavan I don't believe so, so I've added some more changes. This should be all ready after somebody else reviews it.

kidunot89 (approved):

PR #1188, opened 2020-10-05
some connector test cleaned up.

kidunot89:
# Checklist
- Suppress the database error printed to the console when `edd_install()` called.
- Applies the `$this->action_prefix` to all test callback checks.
- All short version arrays `[]` refactor to long version `array()` for consistency


dero (approved):
LGTM, thanks for the improvements, @kidunot89!

PR #1197, opened 2020-10-31
Create a filter for change the stream page menu icon

meetawahab:
Create a filter wp_stream_admin_menu_icon to change the admin menu icon for the stram page and beauty the code.  #hacktoberfest


derekherman:
@meetawahab thank you for contributing a PR, we appreciate you wanting to fix a problem. However, I'm inclined to close the PR because I don't understand the problem you are trying to solve with this filter and I'm not seeing a good reason to add it. Could you please explain how the filter would be used in practice, and with example code someone would use to extend this filter.

PR #1203, opened 2020-12-10
SQL_CALC_FOUND_ROWS statements refactored

### LINKED ISSUE #1200: Remove SQL_CALC_FOUND_ROWS
`SQL_CALC_FOUND_ROWS` is depcreated from MySQL

`The SQL_CALC_FOUND_ROWS query modifier and accompanying FOUND_ROWS() function are deprecated as of MySQL 8.0.17 and will be removed in a future MySQL version`

See https://dev.mysql.com/worklog/task/?id=12615

We need to modify `Query::query()` function to remove `SQL_CALC_FOUND_ROWS` and use `COUNT (*)` to get the total number of rows

```
		/**
		 * BUILD THE FINAL QUERY
		 */
		$query = "SELECT SQL_CALC_FOUND_ROWS {$select}
		FROM $wpdb->stream
		{$join}
		WHERE 1=1 {$where}
		{$orderby}
		{$limits}";

		/**
		 * Filter allows the final query to be modified before execution
		 *
		 * @param string $query
		 * @param array  $args
		 *
		 * @return string
		 */
		$query = apply_filters( 'wp_stream_db_query', $query, $args );

		$result = array();
		/**
		 * QUERY THE DATABASE FOR RESULTS
		 */
		$result['items'] = $wpdb->get_results( $query ); // @codingStandardsIgnoreLine $query already prepared
		$result['count'] = $result['items'] ? absint( $wpdb->get_var( 'SELECT FOUND_ROWS()' ) ) : 0;

		return $result;
```

kidunot89:
Fixes #1200 .

# Checklist
- `SQL_CALC_FOUND_ROWS` removed from MySQL statements.
- Result count assigned a modified count query.

### DIFF
--- a/classes/class-query.php
+++ b/classes/class-query.php
@@ -229,7 +229,7 @@ class Query {
 		/**
 		 * BUILD THE FINAL QUERY
 		 */
-		$query = "SELECT SQL_CALC_FOUND_ROWS {$select}
+		$query = "SELECT {$select}
 		FROM $wpdb->stream
 		{$join}
 		WHERE 1=1 {$where}
@@ -246,12 +246,29 @@ class Query {
 		 */
 		$query = apply_filters( 'wp_stream_db_query', $query, $args );
 
-		$result = array();
+		// Build result count query.
+		$count_query = "SELECT COUNT(*) as found
+		FROM $wpdb->stream
+		{$join}
+		WHERE 1=1 {$where}";
+
+		/**
+		 * Filter allows the result count query to be modified before execution.
+		 *
+		 * @param string $query
+		 * @param array  $args
+		 *
+		 * @return string
+		 */
+		$count_query = apply_filters( 'wp_stream_db_count_query', $count_query, $args );
+
 		/**
 		 * QUERY THE DATABASE FOR RESULTS
 		 */
-		$result['items'] = $wpdb->get_results( $query ); // @codingStandardsIgnoreLine $query already prepared
-		$result['count'] = $result['items'] ? absint( $wpdb->get_var( 'SELECT FOUND_ROWS()' ) ) : 0;
+		$result = array(
+			'items' => $wpdb->get_results( $query ), // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+			'count' => absint( $wpdb->get_var( $count_query ) ), // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+		);
 
 		return $result;
 	}


classes/class-query.php
```diff
@@ -246,12 +246,29 @@ public function query( $args ) {
...
+
+		/**
+		 * Filter allows the result count query to be modified before execution.
+		 *
+		 * @param string $query
+		 * @param array  $args
+		 *
+		 * @return string
+		 */
+		$count_query = apply_filters( 'wp_stream_db_count_query', $count_query, $args );
```
kidunot89, line 263: @ivankruchkoff Here's result count query.

classes/class-query.php
```diff
@@ -246,12 +246,29 @@ public function query( $args ) {
...
+
 		/**
 		 * QUERY THE DATABASE FOR RESULTS
 		 */
-		$result['items'] = $wpdb->get_results( $query ); // @codingStandardsIgnoreLine $query already prepared
-		$result['count'] = $result['items'] ? absint( $wpdb->get_var( 'SELECT FOUND_ROWS()' ) ) : 0;
+		$result = array(
+			'items' => $wpdb->get_results( $query ), // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+			'count' => absint( $wpdb->get_var( $count_query ) ), // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
```
kidunot89, line 270: @ivankruchkoff @kasparsd @derekherman I added explicit ignore comments here. I think don't anymore sanitation needs to be done because all dynamic parts of the query are run thru `$wpdb->prepare()` earlier in the function.

classes/class-query.php
ivankruchkoff, line 263: I don't think we need a filter here.

ivankruchkoff (approved):

classes/class-query.php
Nikschavan: I believe it will be needed, If someone is modifying the select query, they would need to modify the count query as well to make the count match the results shown.

classes/class-query.php
kasparsd: Yeah, all the dynamic parts seem to be fine. However, we're now passing those queries through filters as the very last thing. I guess we need to trust that whoever is using that filter will do their job to sanitize the adjusted parts.

kasparsd (commented):
@kidunot89 Looks like everyone is good with the approach here. Can you please merge this in?

PR #1211, opened 2020-12-23
Buddypress connector implemented

kidunot89:
Partially resolves #1093 

# Summary checklist
- [x] `activity` callbacks tested and passing.
- [x] `group` callbacks tested and passing.
- [x] `option` callbacks tested and passing.
- [x] `xprofile` callbacks tested and passing.
- [x] `callback_xprofile_group_before_save` and `callback_xprofile_field_before_save` callbacks added to the `BuddyPress` connector class to provide a more stable method of capturing the incoming field/field group action that doesn't rely on `$wpdb->insert_id`.

### DIFF
--- a/connectors/class-connector-buddypress.php
+++ b/connectors/class-connector-buddypress.php
@@ -427,6 +427,8 @@ class Connector_BuddyPress extends Connector {
 		$replacement = str_replace( '-', '_', $option );
 
 		if ( method_exists( $this, 'check_' . $replacement ) ) {
+			$method = "check_{$replacement}";
+			$this->{$method}( $old_value, $new_value );
 			call_user_func(
 				array(
 					$this,
--- a/connectors/class-connector-buddypress.php
+++ b/connectors/class-connector-buddypress.php
@@ -61,9 +61,11 @@ class Connector_BuddyPress extends Connector {
 		'groups_unban_member',
 		'groups_remove_member',
 
+		'xprofile_field_before_save',
 		'xprofile_field_after_save',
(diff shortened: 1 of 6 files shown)


connectors/class-connector-buddypress.php
```diff
@@ -1018,15 +1024,34 @@ public function field_action( $field, $action, $meta = array(), $message = null
...
+	 * be logged in the "after_save" callback.
+	 *
+	 * @action xprofile_field_before_save
+	 *
+	 * @param BP_XProfile_Field $field  Field object.
+	 */
+	public function callback_xprofile_field_before_save( $field ) {
+		$this->incoming_field_action = empty( $field->id ) ? 'created' : 'updated';
+	}
+
```
kidunot89, line 1038: @ivankruchkoff Set my comment below

connectors/class-connector-buddypress.php
```diff
@@ -1095,12 +1133,14 @@ public function field_group_action( $group, $action, $meta = array(), $message =
...
-		 * see callback_xprofile_field_after_save for correct behavior.
-		 */
-		$action = ( $group->id === $wpdb->insert_id ) ? 'created' : 'updated';
+		// Bail if the field group's action wasn't captured.
+		if ( is_null( $this->incoming_field_group_action ) ) {
+			return;
+		}
+
+		$action = $this->incoming_field_group_action;
+		unset( $this->incomming_field_group_action );
```
kidunot89, line 1142: @ivankruchkoff I below this is a better way to capture the incoming field action

PR #1214, opened 2020-12-29
Remove duplicated array key FILTER_UNSAFE_RAW

szepeviktor:
Both FILTER_DEFAULT, FILTER_UNSAFE_RAW equal 516.

# Checklist

### DIFF
--- a/classes/class-filter-input.php
+++ b/classes/class-filter-input.php
@@ -35,8 +35,6 @@ class Filter_Input {
 		FILTER_SANITIZE_SPECIAL_CHARS => 'htmlspecialchars',
 		FILTER_SANITIZE_STRING        => 'sanitize_text_field',
 		FILTER_SANITIZE_URL           => 'esc_url_raw',
-		// Other.
-		FILTER_UNSAFE_RAW             => null,
 	);
 
 	/**


kidunot89 (approved):
@szepeviktor It doesn't break anything internally, but could result in a breaking change to custom plugins relying on this functionality in custom connectors. I approve, but that fact should be noted.

szepeviktor:
> custom plugins relying on this functionality

PHP automatically drops the former equal key at declaration.

```
// so
$arr = [
    'pear' => 'fine',
    'apple' => 'good',
    'pear' => 'not-okay',
];
// pear will never be fine
```

classes/class-filter-input.php
```diff
@@ -35,8 +35,6 @@ class Filter_Input {
 		FILTER_SANITIZE_SPECIAL_CHARS => 'htmlspecialchars',
 		FILTER_SANITIZE_STRING        => 'sanitize_text_field',
 		FILTER_SANITIZE_URL           => 'esc_url_raw',
-		// Other.
-		FILTER_UNSAFE_RAW             => null,
```
kasparsd, line 39: Both `FILTER_UNSAFE_RAW` and `FILTER_DEFAULT` are equal to `516` so that makes the last item override the value. However, they are both currently `null` so it doesn't actually change it.

Looks like this data structure would need to be reworked to account for potentially duplicate keys so I suggest we leave it as is until that is done.

kasparsd:
Thank you @szepeviktor for pointing out the issue of the array key override. We probably need to address this in a larger fix that doesn't use keys for matching the PHP filter keys. The current implementation seems to not cause any bugs so I'll close this out for now.

PR #1236, opened 2021-02-22
Settings page defaults repatched

### LINKED ISSUE #1056: Old records are not being removed

kidunot89:
Fixes #1056 again.

Re-patches default settings patch #1159 reverted in #1229 .


codepuncher:
hey @kidunot89, I see that this was scheduled for the `3.7.0` release but was never merged.
Any reason for that?
Quite keen on seeing this one merged 🎉

kidunot89:
@codepuncher I'm not sure. @derekherman @kasparsd ☝🏿

PR #1264, opened 2021-05-10
Posts Connector: Enable on the front end to ensure logging happens for REST API events

coreymckrill:
Post creation and updating that happens through the REST API is not considered to occur in the "admin" even if the request is made from the block editor. Therefore, these events aren't getting logged because the Posts connector is disabled for the frontend. I don't know what the rationale is for disabling it, but by _not_ doing that, post creation gets logged.

Fixes #1195 
Fixes #1250

# Checklist

- [ ] ~Project documentation has been updated to reflect the changes in this pull request, if applicable.~
- [x] I have tested the changes in the local development environment (see `contributing.md`).
- [ ] I have added phpunit tests.

PR #1269, opened 2021-07-08
Connectors: Check the request type and unregister connectors that don't support it

### LINKED ISSUE #1195: Post "update" transition not logged when using Block Editor.
# Bug Report

## Expected Behavior

When updating a post, page or custom post type with the Block Editor I expect the activity to be logged by Stream.

## Actual Behavior

Activity for updates are not getting logged.

Describe what actually happens.

1. Open an existing post for editing with the Block Editor.

2. Make a change to the content (or the title).

3. Click on "Update"

## Steps to Verify the Problem

1. Install and Activate "Classic Editor".

2. Open an existing post for editing with the Block Editor.

3. Make a change to the content (or the title).

4. Click on "Update"

Classic Editor results in the expected behaviour, Block Editor does not.

## System Information

- Stream plugin version: 3.6.0
- WordPress version: 5.5.1
- PHP version: 7.4.1
- Browser: Chrome + Safari
- Computer operating system: MacOS

### LINKED ISSUE #1250: I am not registering the creation of entries or pages
## Expected Behavior

Should record when a user creates content (page, post...)

## Actual Behavior

The log of post and page creation is not reflected.

## System Information

- Stream plugin version: 3.6.2
- WordPress version: 5.7
- PHP version: 4.4.16
- Browser: Firefox
- Computer operating system: Windows 8.1

coreymckrill:
This is an alternative approach to #1264. Since requests to WP APIs are not considered to be `is_admin`, connectors that are disabled for the "frontend" are also disabled for API requests, which is not always (or maybe ever) intended. In this PR we allow all connectors to be registered, regardless of `is_admin`, and then take the following steps when the `parse_request` action is fired (this is when the REST API is initialized):

1. Check to see what kind of request we have, differentiating between admin, frontend, and api requests.
2. Cycle through all the connectors and for each, check if it supports that type of request (via the `register_admin`, `register_frontend`, and the new `register_api` class properties)
3. If a connector does not support the current request type, unregister it.

This addresses the issue of the Posts connector not logging post changes that happen via the REST API, without changing its behavior for actual frontend requests.

Fixes #1195
Fixes #1250

# Checklist

### DIFF
--- a/classes/class-connector.php
+++ b/classes/class-connector.php
@@ -55,6 +55,13 @@ abstract class Connector {
 	public $register_frontend = true;
 
 	/**
+	 * Register connector for API requests
+	 *
+	 * @var bool
+	 */
+	public $register_api = true;
+
+	/**
 	 * Holds connector registration status flag.
 	 *
 	 * @var bool
--- a/classes/class-connectors.php
+++ b/classes/class-connectors.php
@@ -61,6 +61,9 @@ class Connectors {
 	public function __construct( $plugin ) {
 		$this->plugin = $plugin;
 		$this->load_connectors();
+
+		// Run this after the REST API has been initialized.
+		add_action( 'parse_request', array( $this, 'check_request_type' ), 11 );
 	}
 
 	/**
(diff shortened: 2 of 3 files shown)


kasparsd:
This has been fixed by #1264.

PR #1273, opened 2021-08-30
Fixing required/optional parameters

### LINKED ISSUE #1272: PHP 8 - PHP Deprecated:  Required parameter $alert follows optional parameter $context
# Bug Report

PHP 8 is deprecating required function/method parameters following optional parameters. The `Alert_Trigger` abstract class and all four child classes violate this by this signature:

```php
public function get_display_value( $context = 'normal', $alert ){}
```

Which raises warnings such as:

```
PHP Deprecated:  Required parameter $alert follows optional parameter $context in stream/alerts/class-alert-trigger-author.php on line 161
```

This was mentioned in a comment attached to #1202 but I thought it would be best to have a dedicated ticket since this is pretty easy to fix.

cjhaas:
Fixes #1272 .

This could be fixed in one of two ways, either just update the second parameter to have `null` as the default value, or to update all logic to flop the two parameters.

This PR does the former since it shouldn't have any BC issues. All uses of `$alert` inside of the methods appear to be used to bring out additional meta data, and the existing code uses an `empty()` check which is [safe with `null` parameters](https://3v4l.org/YZoX3)

It is possible that anyone else that has custom code with additional alerts, however, might run into signature issues depending on their PHP versions.

# Checklist

- [ ] ~Project documentation has been updated to reflect the changes in this pull request, if applicable.~
- [ ] ~I have added phpunit tests.~

## Release Changelog

- Fix: Fixes PHP 8 deprecation warning

## Release Checklist

### DIFF
--- a/alerts/class-alert-trigger-action.php
+++ b/alerts/class-alert-trigger-action.php
@@ -164,12 +164,12 @@ class Alert_Trigger_Action extends Alert_Trigger {
 	 *
 	 * @see Alert_Trigger::get_display_value().
 	 *
-	 * @param string $context The location this data will be displayed in.
-	 * @param Alert  $alert Alert being processed.
+	 * @param string     $context The location this data will be displayed in.
+	 * @param Alert|null $alert Alert being processed.
 	 *
 	 * @return string
 	 */
-	public function get_display_value( $context = 'normal', $alert ) {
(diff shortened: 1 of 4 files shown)


kasparsd (approved):
This looks great! Thanks for the contribution @cjhaas!

PR #1285, opened 2021-09-16
Create the database tables if they do not exist when resetting the DB

### LINKED ISSUE #1284: Stream does not check database tables exist
# Bug Report

Stream only creates database tables on plugin activation, then never checks they exist.

## Expected Behavior

Use Stream in an environment here plugin activation occurs.

Stream sees that the tables don't exist and creates them.

## Actual Behavior

Stream does not create the tables because you did not load the plugin by activating it.

Additionally, the reset database tables link in settings does not create the tables if they do not exist

1. Install Stream using a method that does not involve plugin activation. E.g. as an MU Plugin
2. Try to use it
3. Be constantly told that the database tables do not exist

## System Information

- Stream plugin version:
- WordPress version: 5.8
- PHP version: 7.4
- Browser: Chrome
- Computer operating system: Linux

If making it auto-check isn't possible, providing a way to create the tables would also be great. Right now this is extremely frustrating and regularly causes problems with local environments on multiple projects.

tomjn:
Create the database tables if they do not exist when resetting the database

Fixes #1284

If the plugin is loaded without activation, tables are never created and resetting the database does not create tables. This PR changes that so that if tables are missing, a reset creates them.

# Checklist

## Release Changelog

## Release Checklist


kasparsd (commented):
@tomjn Would this still be necessary if we merge #1286 which creates the tables when the admin loads?

tomjn:
@kasparsd I don't believe it's strictly necessary, but it is nice and makes sense that a DB reset would make sure the tables are present and up to date. This is explicit where the other PR is implicit.

lkraav:
> I don't believe it's strictly necessary, but it is nice and makes sense that a DB reset would make sure the tables are present and up to date. This is explicit where the other PR is implicit.

I've been wondering if x+1 plugins with custom tables all do their "TABLE EXISTS" checks, does it have any real performance impact or not.

In the big picture, it does seem kind of redundant to guard against a real far-edge case millions of times / year * # of plugins.

tomjn:
@lkraav for this PR it'll happen when a user clicks a button, as for WP Admin or CLI it's pretty safe. On the frontend it can add up quickly on a busy site

classes/class-admin.php
```diff
@@ -673,6 +673,9 @@ public function wp_ajax_reset() {
 				esc_html__( "You don't have sufficient privileges to do this action.", 'stream' )
 			);
 		}
+		
+		// Create and update the database tables if they do not exist.
+		$this->plugin->install->check();
```
kasparsd, line 678: Is this check here `defined( 'DOING_AJAX' ) && DOING_AJAX` going to stop this particular call from doing anything?

https://github.com/tomjn/stream/blob/02baea3f526407a0883bee1ccf769562553d012e/classes/class-install.php#L91-L93

tomjn: 🤦🏻 I wouldn't have noticed without your help :)

tomjn: Do you think `$this->plugin->install->install( $this->plugin->get_version() );` might make more sense?

PatelUtkarsh: Yes, that works. @tomjn Let me know if you'd like to patch or I can take over.

tomjn: @PatelUtkarsh it's been years since I looked at this, happy to hand over to you

PR #1289, opened 2021-10-06
Limit the ORDER value to only ASC or DESC

kasparsd:
Limit the record query ordering to `ASC` or `DESC` only.

# Checklist

### DIFF
--- a/classes/class-query.php
+++ b/classes/class-query.php
@@ -179,21 +179,30 @@ class Query {
 		/**
 		 * PARSE ORDER PARAMS
 		 */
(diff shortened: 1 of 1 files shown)


classes/class-query.php
```diff
@@ -179,21 +179,30 @@ public function query( $args ) {
 		/**
 		 * PARSE ORDER PARAMS
 		 */
-		$order     = esc_sql( $args['order'] );
```
kasparsd, line 182: Order was allowed to be anything.

PR #1309, opened 2022-03-08
Update deps and WPCS before release

kasparsd:
- Update Node and Composer deps.
- WPCS formatting prior release.
- Switch to MariaDB containers to make it compatible with ARM Macs.

### DIFF
--- a/tests/tests/connectors/test-class-connector-edd.php
+++ b/tests/tests/connectors/test-class-connector-edd.php
@@ -18,9 +18,6 @@ class Test_WP_Stream_Connector_EDD extends WP_StreamTestCase {
 
 		$this->plugin->connectors->unload_connectors();
 
-		$post_connector = new Connector_Posts();
-		$post_connector->register();
-
 		// Make partial of Connector_EDD class, with mocked "log" function.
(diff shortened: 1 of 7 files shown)


kasparsd:
Looks like #1264 made [one of the unit tests](https://app.travis-ci.com/github/xwp/stream/jobs/562444098#L1671) for EDD fail:

> 1) WP_Stream\Test_WP_Stream_Connector_EDD::test_log_override
Failed asserting that 2 is identical to 4.

https://github.com/xwp/stream/blob/305583d70e8a7667d1f99df85196024a7334ed50/tests/tests/connectors/test-class-connector-edd.php#L172-L193

I'm not sure why the `wp_stream_log_data` is called twice for every post insert.

tests/tests/connectors/test-class-connector-edd.php
```diff
@@ -18,9 +18,6 @@ public function setUp() {
 
 		$this->plugin->connectors->unload_connectors();
 
-		$post_connector = new Connector_Posts();
```
kasparsd, line 21: This was causing the `wp_stream_log_data` action to be called twice -- once for the EDD connector and once for the Posts connector and the test in this file was being run during both.

PR #1311, opened 2022-03-08
Set min. PHP version in Composer config

szepeviktor:
Just that.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,6 +11,7 @@
     }
   ],
   "require": {
+    "php": "^5.3",
     "composer/installers": "~1.0"
   },
   "require-dev": {


composer.json
```diff
@@ -11,6 +11,7 @@
     }
   ],
   "require": {
+    "php": "^5.3",
```
kasparsd, line 14: The platform PHP further down the file is set to `5.6.20` which matches the minimum version of WP core https://wordpress.org/about/requirements/

> WordPress also works with PHP 5.6.20+ and MySQL 5.0+, but these versions have reached official End Of Life and as such may expose your site to security vulnerabilities.

Should we set this to `>=5.6.20` then?

The goal is to support every PHP from 5.6.20 to 8.x.

szepeviktor: I'm not familiar with this project.
https://github.com/xwp/stream/blob/1b308298417274b0030c0ecc6ceec5dbb33e27a2/stream.php#L34
This line contains 5.3

Please decide what to do!
A month ago my default moved to PHP 8.0

szepeviktor: Friendly ping 🏓

szepeviktor:
Done in 78d5369

PR #1327, opened 2022-04-06
Prevent Woocommerce context to be autoselected after filtering

jmslbam:
Check if $page_id has a proper value, before adding it to the list. Because adding an empty key to this array results in "Woocommerce" always being selected after a filter action.

Screenshot 2022-04-06 at 12 37 29

This happend when a third party plugin added a Woocommerce Settingpage without properly setting the `page_id`. Woocommerce doesn't give you a error when doing this. But this does result in a empty `key` in the `$context_labels` https://github.com/xwp/stream/blob/develop/connectors/class-connector-woocommerce.php#L177 here.

The settingspage with the empty `page_id` is retreived from `\WC_Admin_Settings::get_settings_pages()` on https://github.com/xwp/stream/blob/develop/connectors/class-connector-woocommerce.php#L768. So it's a WC thing, but regresses into Stream.

I found the solution by reading this issue #1051

Before diving into PHPUnit etc (not fluently with them) I wanted to know if this would be a PR that would get accepted.

Kind regards,

Jaime

# Checklist

## Release Changelog

- Fix: Woocommerce context dropdown filter being selected by default after a "filter action"
- New: Woocommerce filter doesn't get selected by default anymore.

## Release Checklist

### DIFF
--- a/connectors/class-connector-woocommerce.php
+++ b/connectors/class-connector-woocommerce.php
@@ -778,6 +778,10 @@ class Connector_Woocommerce extends Connector {
 				if ( empty( $sections ) ) {
 					$sections[''] = $page_label;
 				}
+				
+				if( empty( $page_id ) ) {
+				 	continue;
+				}
 
 				$settings_pages[ $page_id ] = $page_label;
 


jmslbam:
I meant #1051 when referring to a simliar situation :)

jmslbam:
I fixed it via 


```
/**
	 * Drop empty WooCommerce context keys so Stream does not auto-select
	 * the WooCommerce connector when filtering (e.g. by user role).
	 *
	 * Caused by a WC settings page registered without a page_id, which
	 * collides with Stream's empty "show all contexts" option.
	 *
	 * @see https://github.com/xwp/stream/pull/1327
	 */
	public function remove_empty_woocommerce_contexts( $contexts ) {
		if ( ! is_array( $contexts ) ) {
			return $contexts;
		}

		return array_filter(
			$contexts,
			static function ( $key ) {
				return '' !== (string) $key;
			},
			ARRAY_FILTER_USE_KEY
		);
	}
```

Instead of the PR which checked on page_id

```
if ( empty( $page_id ) ) {
	continue;
}
```

PR #1355, opened 2022-08-23
Fixes #1294 widget connector php 8 incompatibility

### LINKED ISSUE #1294: [PHP 8] Error when updating widgets.
# Bug Report
While Stream is active on PHP 8 host, adding or removing widgets in WordPress causes an error to be thrown.  The named arguments added to PHP 8 create an issue with the call chain found at: `/stream/connectors/class-connector-widgets.php` lines 300, 301, as well as 360 and 361

## Expected Behavior
The expected behavior is simply saving the widgets without delay or error messages.

## Actual Behavior
When saving, update may take a moment and then report there was an error via pop-up at the bottom of the screen, which you can easily glance over if you're scrolling about.   
If you have logging enabled, you will find the following message:  
`PHP Fatal error:  Uncaught ArgumentCountError: array_merge() does not accept unknown named parameters in /var/www/dcr/wp-content/plugins/stream/connectors/class-connector-widgets.php:300`

1. Use PHP 8 with latest WordPress 5.8.1
2. Install latest Stream plugin
3. Create a couple widgets and then delete one to trigger the functions.
4. Check your error logging for details.

## System Information

- Stream plugin version: 3.8.1
- WordPress version: 5.8.1
- PHP version: 8.0.11

## Possible Solution

If you're using PHP 8, you can patch the error-prone lines by adding the parameter names to `call_user_func_array` so that lines 300, 301, as well as 360 and 361 look like this:
```php
$all_old_widget_ids = array_unique( call_user_func_array( callback:'array_merge', args:$old ) );
$all_new_widget_ids = array_unique( call_user_func_array( callback:'array_merge', args:$new ) );
```
The issue with this patch is that it clearly is not syntactically compatible with any previous version of PHP.  I'm sure there is a more compatible way to patch this, I just don't have the time to figure it out at the moment.

ParhamG:
Fixes #1294

The widget connector throws PHP fatal errors on widget add/remove on PHP 8.x:
`PHP Fatal error:  Uncaught ArgumentCountError: array_merge() does not accept unknown named parameters in /wp-content/plugins/stream/connectors/class-connector-widgets.php:300`

This update fixes the issue by applying `array_values()` on the arguments before passing them to `array_merge()/call_user_func_array()`.

# Checklist

- [X] I have tested the changes in the local development environment (see `contributing.md`).

### DIFF
--- a/connectors/class-connector-widgets.php
+++ b/connectors/class-connector-widgets.php
@@ -297,8 +297,8 @@ class Connector_Widgets extends Connector {
 	 * @return void
 	 */
 	protected function handle_widget_removal( $old, $new ) {
-		$all_old_widget_ids = array_unique( call_user_func_array( 'array_merge', $old ) );
-		$all_new_widget_ids = array_unique( call_user_func_array( 'array_merge', $new ) );
+		$all_old_widget_ids = array_unique( call_user_func_array( 'array_merge', array_values( $old ) ) );
+		$all_new_widget_ids = array_unique( call_user_func_array( 'array_merge', array_values( $new ) ) );
 		// @todo In the customizer, moving widgets to other sidebars is problematic because each sidebar is registered as a separate setting; so we need to make sure that all $_POST['customized'] are applied?
 		// @todo The widget option is getting updated before the sidebars_widgets are updated, so we need to hook into the option update to try to cache any deletions for future lookup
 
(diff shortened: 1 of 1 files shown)


kasparsd (approved):
Thank you for the pull request @ParhamG! 

Looks like several users experiencing the same issue have tested the patch and reported that it works fine.

PR #1365, opened 2022-09-13
Optionsfix

### LINKED ISSUE #1056: Old records are not being removed

timbutler:
Potentially fixes #1056.

We discovered an issue where some defaults were saved to the database but not the complete set of fields. This meant that unless fields such as `general_records_ttl` were set, it wouldn't load the defaults and therefore never purge the records. The issue has the same description as #1056 but unsure of the exact cause from the bug reporter.

When reviewing the code, I could see that the Settings class already had the correct syntax for merging defaults so it was easiest to therefore adjust the Admin class to simply use this (instead of duplicating). By ensuring the full defaults are loaded (if not set in the database), this also ensures any new options added won't be dropped in the future. 

This may also resolve #1236.

### DIFF
--- a/classes/class-admin.php
+++ b/classes/class-admin.php
@@ -143,9 +143,7 @@ class Admin {
 		add_filter( 'role_has_cap', array( $this, 'filter_role_caps' ), 10, 3 );
 
 		if ( is_multisite() && $plugin->is_network_activated() && ! is_network_admin() ) {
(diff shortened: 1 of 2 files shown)


timbutler:
Sorry, tried to follow the contributions guide for which branch but because they're out-of-sync it's not going to be correct for develop. If required, I can start with the develop branch and push the changes there.

PR #1391, opened 2023-01-09
Check authorization on 'save_new_alert' AJAX action

schlessera:
The logic in `save_new_alert` AJAX action does only check for a valid nonce, but not for authorization. This makes it possible to reuse a valid nonce and trigger the `save_new_alert` with an unauthorized or unauthenticated user.

This PR adds an authorization check to `save_new_alert` (as well as to `get_new_alert_triggers_notifications`, which could be used to retrieve a nonce as an authenticated but unauthorized user), and adds corresponding tests to ensure both wanted and unwanted requests behave as expected with regards to alert creation.

# Checklist

### DIFF
--- a/classes/class-alerts.php
+++ b/classes/class-alerts.php
@@ -509,6 +509,7 @@ class Alerts {
 	 * @return void
 	 */
 	public function display_notification_box( $post = array() ) {
+		$alert      = null;
 		$alert_type = 'none';
 		if ( is_object( $post ) ) {
 			$alert      = $this->get_alert( $post->ID );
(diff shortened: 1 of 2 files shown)


classes/class-alerts.php
```diff
@@ -509,6 +509,7 @@ public function change_menu_link_url() {
 	 * @return void
 	 */
 	public function display_notification_box( $post = array() ) {
+		$alert      = null;
```
schlessera, line 512: This was added to avoid PHP notices for the use of a non-initialized variable.

kasparsd (approved):
This looks great, thanks!

PR #1431, opened 2023-04-20
Add missing nonce check to uninstall action

### LINKED ISSUE #1426: Vulnerability in 3.9.2
# Feature Request / Question

Patchstack reported version 3.9.2 as being vulnerable to CSRF 
[https://patchstack.com/database/vulnerability/stream/wordpress-stream-plugin-3-9-2-cross-site-request-forgery-csrf-vulnerability?_a_id=431](https://patchstack.com/database/vulnerability/stream/wordpress-stream-plugin-3-9-2-cross-site-request-forgery-csrf-vulnerability?_a_id=431)

Any chance you have it in your queue?

schlessera:
Fixes #1426.

This PR includes the following changes:

* adds tests to trigger and assert the vulnerability described in [CVE-2022-43490](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-43490), props [@Lucisu](https://github.com/Lucisu) via [Patchstack](https://patchstack.com/)
* fixes [CVE-2022-43490](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-43490) by adding a nonce check to the `'wp_ajax_wp_stream_uninstall'` AJAX action.
* fixes a broken SQL query that failed to delete user meta records on multisite installations.

# Checklist

### DIFF
--- a/classes/class-db-driver-wpdb.php
+++ b/classes/class-db-driver-wpdb.php
@@ -173,6 +173,20 @@ class DB_Driver_WPDB implements DB_Driver {
 	 * @return \WP_Stream\Uninstall
 	 */
 	public function purge_storage( $plugin ) {
+		// Avoid duplicate uninstall action.
+		// TODO: This is nasty code right now due to the fact the object instance is only known to the action.
+		// The architecture around how the uninstall action is registered should be rethought.
+		if ( has_action( 'wp_ajax_wp_stream_uninstall' ) ) {
+			// Get a list of all callbacks registered to the default priority 10 for the uninstall action.
+			$callbacks = array_values( $GLOBALS['wp_filter']['wp_ajax_wp_stream_uninstall']->callbacks[10] );
+			// From that list, retrieve function of the first (and only) callback.
+			$callback_function = $callbacks[0]['function'];
+			// The callback's first element should be the object instance of the Uninstall class.
+			$uninstall = $callback_function[0];
+
+			return $uninstall;
+		}
+
 		$uninstall = new Uninstall( $plugin );
 		add_action( 'wp_ajax_wp_stream_uninstall', array( $uninstall, 'uninstall' ) );
 
--- a/classes/class-uninstall.php
+++ b/classes/class-uninstall.php
@@ -52,6 +52,8 @@ class Uninstall {
 	 * Uninstall Stream by deleting its data
 	 */
 	public function uninstall() {
+		check_ajax_referer( 'stream_uninstall_nonce', 'nonce' );
+
 		$this->options = array(
 			$this->plugin->install->option_key,
 			$this->plugin->settings->option_key,
@@ -235,6 +237,10 @@ class Uninstall {
 	 */
 	private function deactivate() {
 		deactivate_plugins( $this->plugin->locations['plugin'] );
+
+		if ( defined( 'WP_STREAM_TESTS' ) && WP_STREAM_TESTS ) {
+			return true;
+		}
 
 		wp_safe_redirect(
 			add_query_arg(
--- a/classes/class-uninstall.php
+++ b/classes/class-uninstall.php
@@ -220,7 +220,7 @@ class Uninstall {
 		// Specific user meta.
 		foreach ( $this->user_meta as $meta_key ) {
 			$wpdb->query(
-				$wpdb->prepare( "DELETE FROM {$wpdb->usermeta} WHERE meta_key = {$wpdb->prefix}%s;", $meta_key )
+				$wpdb->prepare( "DELETE FROM {$wpdb->usermeta} WHERE meta_key = %s;", $wpdb->prefix . $meta_key )
 			);
 		}
 	}
--- a/tests/tests/test-class-admin.php
+++ b/tests/tests/test-class-admin.php
@@ -392,7 +392,7 @@ class Test_Admin extends WP_StreamTestCase {
 
 		try {
 			$this->_handleAjax( 'wp_stream_filters' );
-		} catch ( WPAjaxDieStopException $e ) {}
+		} catch ( \WPAjaxDieStopException $e ) {}
 
 		// Check that the exception was thrown.
 		$this->assertTrue( isset( $e ) );
--- a/tests/tests/test-class-alerts.php
+++ b/tests/tests/test-class-alerts.php
@@ -382,22 +382,15 @@ class Test_Alerts extends WP_StreamTestCase {
 		wp_set_current_user( $user_id );
 
 		$alerts = new Alerts( $this->plugin );
-		try {
-			$_POST['wp_stream_trigger_author'] = 'me';
-			$_POST['wp_stream_trigger_context'] = 'posts-post';
-			$_POST['wp_stream_trigger_action'] = 'edit';
-			$_POST['wp_stream_alert_type'] = 'highlight';
 
-			$this->expectException( 'WPAjaxDieStopException' );
-			$this->_handleAjax( 'save_new_alert' );
-		} catch ( \WPAjaxDieContinueException $e ) {
-			$exception = $e;
-			// Check that the exception was thrown.
-			$this->assertTrue( isset( $exception ) );
+		$_POST['wp_stream_trigger_author'] = 'me';
+		$_POST['wp_stream_trigger_context'] = 'posts-post';
+		$_POST['wp_stream_trigger_action'] = 'edit';
+		$_POST['wp_stream_alert_type'] = 'highlight';
 
-			// The output should be a -1 for failure.
-			$this->assertEquals( '-1', $exception->getMessage() );
-		}
+		$this->expectException( 'WPAjaxDieStopException' );
+		$this->expectExceptionMessage( '-1' );
+		$this->_handleAjax( 'save_new_alert' );
 	}
 
 	/**
@@ -409,23 +402,16 @@ class Test_Alerts extends WP_StreamTestCase {
 		wp_set_current_user( $user_id );
 
 		$alerts = new Alerts( $this->plugin );
-		try {
-			$_POST['wp_stream_trigger_author'] = 'me';
-			$_POST['wp_stream_trigger_context'] = 'posts-post';
-			$_POST['wp_stream_trigger_action'] = 'edit';
-			$_POST['wp_stream_alert_type'] = 'highlight';
-			$_POST['wp_stream_alerts_nonce'] = 'invalid';
 
-			$this->expectException( 'WPAjaxDieStopException' );
-			$this->_handleAjax( 'save_new_alert' );
-		} catch ( \WPAjaxDieContinueException $e ) {
-			$exception = $e;
-			// Check that the exception was thrown.
-			$this->assertTrue( isset( $exception ) );
+		$_POST['wp_stream_trigger_author'] = 'me';
+		$_POST['wp_stream_trigger_context'] = 'posts-post';
+		$_POST['wp_stream_trigger_action'] = 'edit';
+		$_POST['wp_stream_alert_type'] = 'highlight';
+		$_POST['wp_stream_alerts_nonce'] = 'invalid';
 
-			// The output should be a -1 for failure.
-			$this->assertEquals( '-1', $exception->getMessage() );
-		}
+		$this->expectException( 'WPAjaxDieStopException' );
+		$this->expectExceptionMessage( '-1' );
+		$this->_handleAjax( 'save_new_alert' );
 	}
 
 	/**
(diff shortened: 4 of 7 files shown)


classes/class-db-driver-wpdb.php
```diff
@@ -173,6 +173,20 @@ public function setup_storage( $plugin ) {
...
+			// Get a list of all callbacks registered to the default priority 10 for the uninstall action.
+			$callbacks = array_values( $GLOBALS['wp_filter']['wp_ajax_wp_stream_uninstall']->callbacks[10] );
+			// From that list, retrieve function of the first (and only) callback.
+			$callback_function = $callbacks[0]['function'];
+			// The callback's first element should be the object instance of the Uninstall class.
+			$uninstall = $callback_function[0];
+
+			return $uninstall;
+		}
+
```
schlessera, line 189: @kasparsd During testing, the `purge_storage()` method is sometimes triggered twice. Any suggestions on how to deal with this in a reliable without rewriting parts of the architecture?
Do you think prefixing this hack with a check on `WP_STREAM_TESTS` would be enough for now?

classes/class-uninstall.php
```diff
@@ -52,6 +52,8 @@ public function __construct( $plugin ) {
 	 * Uninstall Stream by deleting its data
 	 */
 	public function uninstall() {
+		check_ajax_referer( 'stream_uninstall_nonce', 'nonce' );
```
schlessera, line 55: This is the actual security fix. Most of the rest of the PR is about reliable testing to make sure the fix does what it is supposed to do.

classes/class-uninstall.php
```diff
@@ -218,7 +220,7 @@ private function delete_blog_user_meta( $blog_id = 1 ) {
 		// Specific user meta.
 		foreach ( $this->user_meta as $meta_key ) {
 			$wpdb->query(
-				$wpdb->prepare( "DELETE FROM {$wpdb->usermeta} WHERE meta_key = {$wpdb->prefix}%s;", $meta_key )
+				$wpdb->prepare( "DELETE FROM {$wpdb->usermeta} WHERE meta_key = %s;", $wpdb->prefix . $meta_key )
```
schlessera, line 223: This is a bug I detected while adding tests to multisite uninstallation. That probably never worked as intended, it generates a meta key like `stream'my_meta-key'` (prefix outside of quotes).

tests/tests/test-class-alerts.php
```diff
@@ -382,22 +382,15 @@ function test_save_new_alert_no_nonce() {
...
-			$_POST['wp_stream_trigger_action'] = 'edit';
-			$_POST['wp_stream_alert_type'] = 'highlight';
 
-			$this->expectException( 'WPAjaxDieStopException' );
-			$this->_handleAjax( 'save_new_alert' );
-		} catch ( \WPAjaxDieContinueException $e ) {
-			$exception = $e;
-			// Check that the exception was thrown.
-			$this->assertTrue( isset( $exception ) );
+		$_POST['wp_stream_trigger_author'] = 'me';
```
schlessera, line 386: I simplified some of these tests as they did a weird combination of expecting an exception but also intercepting it before it is shown to PHPUnit.

tests/tests/test-class-db-driver-wpdb.php
```diff
@@ -98,6 +111,190 @@ public function test_table_names() {
...
+		// Test cases override table dropping queries to prevent them
+		// from being executed, but we need actual dropping here.
+		foreach ( $GLOBALS['wp_filter']['query'] as $priority => $filters ) {
+			foreach ( $filters as $filter => $data ) {
+				if ( false !== strpos( $filter, '_drop_temporary_table' ) ) {
+					unset( $filters[$filter] );
+					$GLOBALS['wp_filter']['query'][$priority] = $filters;
+				}
+			}
+		}
```
schlessera, line 131: This is needed because our `dbdelta` SQL is using `CREATE TABLE`, but the drops are turned into `DROP TEMPORARY TABLE` by default by the WP test case, breaking the drops as the temporary tables were never created (they were regular tables instead).

tests/tests/test-class-db-driver-wpdb.php
```diff
@@ -98,6 +111,190 @@ public function test_table_names() {
...
+
+		$stream_count = $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(*) FROM {$this->driver->table} WHERE site_id = %d", $blog_id ) );
+		$this->assertEquals( 0, $stream_count );
+		$stream_id_placeholders = implode( ', ', array_fill( 0, count( $stream_ids ), '%d' ) );
+		$stream_meta_count = $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(*) FROM {$this->driver->table_meta} WHERE record_id IN ( $stream_id_placeholders )", $stream_ids ) );
+		$this->assertEquals( 0, $stream_meta_count );
+	}
+
+	// Test that the purge_storage() function can only be called with admin capabilities.
+	// TODO: This throws a notice: "Test code or tested code did not (only) close its own output buffers".
```
schlessera, line 220: The tests that throw an exception here are flagged as `risky` because they seem to break the output buffer logic somewhere. However, I have failed to pinpoint the cause of this as of yet.

classes/class-db-driver-wpdb.php
```diff
@@ -170,13 +170,10 @@ public function setup_storage( $plugin ) {
...
 	 *
 	 * @param \WP_Stream\Plugin $plugin Instance of the plugin.
-	 * @return \WP_Stream\Uninstall
 	 */
 	public function purge_storage( $plugin ) {
 		$uninstall = new Uninstall( $plugin );
-		add_action( 'wp_ajax_wp_stream_uninstall', array( $uninstall, 'uninstall' ) );
-
-		return $uninstall;
+		$uninstall->uninstall();
```
kasparsd, line 176: We expect that calling `purge_storage()` will actually trigger all the purge steps so we call the `uninstall()` logic here.

We also expected that whoever is calling the `purge_storage()` method has done all the necessary input validation and permission checks. This decouples the routing knowledge from the actual business logic of purging the data.

tests/bootstrap.php
```diff
@@ -67,12 +67,12 @@ function xwp_install_edd() {
 
 	edd_install();
 
-	global $current_user, $edd_options;
+	global $edd_options;
 
 	$edd_options = get_option( 'edd_settings' );
 
-	$current_user = new WP_User(1);
```
kasparsd, line 74: This was setting up the test environment as if an admin user had logged in. It should be the responsibility of each test to setup and tear-down the environment state as necessary.

tests/tests/connectors/test-class-connector-editor.php
```diff
@@ -14,10 +14,6 @@ public function setUp() {
 		$this->mock->register();
 	}
 
-	public function tearDown() {
```
kasparsd, line 17: These are not needed since they call the parent with no additions.

tests/tests/test-class-admin.php
```diff
@@ -13,18 +13,10 @@ public function setUp() {
...
-		$administrator_id = $this->factory->user->create(
-			array(
-				'role'       => 'administrator',
-				'user_login' => 'test_admin',
-				'email'      => 'test@land.com',
-			)
-		);
-		wp_set_current_user( $administrator_id );
+	public function test_logged_out_by_default() {
+		$this->assertFalse( is_user_logged_in(), 'Run tests as logged-out user by default' );
```
kasparsd, line 19: Ensure that default tests run without having a logged-in user.

tests/tests/test-class-admin.php
```diff
@@ -135,6 +127,8 @@ public function test_register_menu() {
 		global $menu;
 		$menu = array(); //phpcs override okay
 
+		$this->login_with_role( 'administrator' );
```
kasparsd, line 130: Now login only when necessary and with the expected role/capabilities.

tests/tests/test-class-admin.php
```diff
@@ -380,26 +377,18 @@ public function test_filter_role_caps() {
 	 * @requires PHPUnit 5.7
 	 */
 	public function test_ajax_filters() {
-		$user = new \WP_User( get_current_user_id() );
-
 		$this->_setRole( 'subscriber' );
 
+		$user = new \WP_User( get_current_user_id() );
```
kasparsd, line 382: The `$this->_setRole()` creates a new user with the required role and authenticates it so we should call `get_current_user_id()` only after it has been done.

tests/tests/test-class-admin.php
```diff
@@ -380,26 +377,18 @@ public function test_filter_role_caps() {
...
+
 		$_POST['filter'] = 'user_id';
 		$_POST['q'] = $user->display_name;
 		$_POST['nonce'] = wp_create_nonce( 'stream_filters_user_search_nonce' );
 
 		$this->expectException( 'WPAjaxDieStopException' );
+		$this->expectExceptionMessage( '-1' );
```
kasparsd, line 389: This is a simplification for the lines above to make it match the rest of the tests.

tests/tests/test-class-admin.php
```diff
@@ -427,29 +416,94 @@ public function test_get_users_record_meta() {
...
 
-		$this->admin->delete_user_meta( get_current_user_id(), $key );
+		$this->expectException( 'WPAjaxDieStopException' );
+		$this->expectExceptionMessage( 'You don&#039;t have sufficient privileges to do this action.' );
 
-		$this->assertEmpty( get_user_meta( get_current_user_id(), $key, true ) );
+		$this->_handleAjax( 'wp_stream_uninstall' );
+	}
+
+	public function test_action_wp_stream_uninstall_not_available_to_non_admins() {
```
kasparsd, line 487: All the tests related to the `stream_uninstall_nonce` action are now part of the admin test suite since that is where the controller logic for this lives.

classes/class-db-driver-wpdb.php
kasparsd: This was happening because every test calling `purge_storage()` was creating a new instance of `Install()` which was added as new `wp_ajax_wp_stream_uninstall` action callback which then all got called when we triggered the action.

```php
add_action( 'aaa', [ new Uninstall( $this->plugin ), 'uninstall' ] );
add_action( 'aaa', [ new Uninstall( $this->plugin ), 'uninstall' ] );
```

results in:

```php
/var/www/html/wp-content/plugins/stream-src/tests/tests/test-class-db-driver-wpdb.php:33:
class WP_Hook#43 (5) {
  public $callbacks =>
  array(1) {
    [10] =>
    array(2) {
      '000000000a8a85d000000000584fc7efuninstall' =>
      array(2) {
        ...
      }
      '000000000a8a85d600000000584fc7efuninstall' =>
      array(2) {
        ...
      }
    }
  }
```
```

File: tests/tests/test-class-db-driver-wpdb.php (hunk 6 above)
Comment (kasparsd) [reply tests/tests/test-class-db-driver-wpdb.php]: This was due to the `catch()` wrappers which are not needed for the `_handleAjax()` helper and the associated logic that relies on catching the `dieHandler` override here:

https://github.com/wp-phpunit/wp-phpunit/blob/a21378376bbec9366c4b18a77d62f42a7dfdabac/includes/testcase-ajax.php#L216-L228

File: classes/class-uninstall.php [hunk 15]
```diff
@@ -58,21 +58,8 @@ public function uninstall() {
...
-
-		if ( defined( 'DISALLOW_FILE_MODS' ) && true === DISALLOW_FILE_MODS ) {
-			wp_die(
-				esc_html__( "You don't have sufficient file permissions to do this action.", 'stream' )
-			);
-		}
-
 		// Prevent this action from firing.
-		remove_action( 'deactivate_plugin', array( 'Connector_Installer', 'callback' ), null );
+		remove_action( 'deactivate_plugin', array( 'WP_Stream\Connector_Installer', 'callback' ), null );
```
kasparsd, line 62: This was leaving a deactivation log in the DB for the blog where the plugin was being uninstalled because of the missing fully-qualified namespace.

classes/class-uninstall.php
```diff
@@ -81,9 +68,10 @@ public function uninstall() {
 
 		/**
 		 * Drop everything on single site installs or when network activated
-		 * Otherwise only delete data relative to the current blog.
+		 * AND being uninstalled from the main blog. Otherwise only delete
+		 * data relative to the current blog.
 		 */
-		if ( ! is_multisite() || $this->plugin->is_network_activated() ) {
+		if ( ! is_multisite() || ( $this->plugin->is_network_activated() && is_main_site() && is_super_admin() ) ) {
```
kasparsd, line 74: @schlessera This fixes a really important issue where the plugin being uninstalled from a blog on a multisite would delete ALL Stream data, if the Stream plugin was activated at the network level.

We now purge all plugin records and tables only:

1. if the uninstall is happening on a single site, or
2. if the uninstall is happening on the main site of the multisite and is triggered by the super admin.

kasparsd (approved):
I've tested all possible scenarios with the plugin Uninstall on a network and a single site, and everything appears to be working.

@schlessera Could you please review my comments inline the pull request and the major issue I discovered with multisite uninstalls. Thanks!

classes/class-uninstall.php
schlessera: Hmm, doesn't that mean that it won't clean up anymore if you only installed it on a subsite?

tests/tests/test-class-admin.php
```diff
@@ -475,4 +529,24 @@ private function dummy_meta_data( $stream_id ) {
...
+	 *
+	 * @param  string $role Use role.
+	 * @return int User ID.
+	 */
+	private function login_with_role( $role ) {
+		$user_id = $this->factory->user->create(
+			array(
+				'role'       => $role,
+				'user_login' => 'test_admin',
+				'email'      => 'test@land.com',
```
schlessera, line 544: In case one tests uses multiple roles, you don't want it to run into errors for non-unique emails and such.
```suggestion
				'user_login' => "test_{$role}}",
				'email'      => "test_{$role}@land.com",
```

schlessera (commented):
One minor change request to ensure different roles can be used at the same time during tests for the role helper.

classes/class-uninstall.php
kasparsd: That's a great idea! Will change that up.

classes/class-uninstall.php
kasparsd: The Stream tables are shared between all sites on a multisite.

Currently the tables are completely removed only if:

1. The plugin is "Uninstalled" from the main blog of the multisite _and_ it was enabled at the network level.
2. The plugin is "Uninstalled" from a non-multisite setup.

All other uninstalls simply delete the data associated with the particular `blog_id`. So in case of uninstalling the plugin from a blog on a network (when the plugin is not activated globally) it would preserve the tables and remove the data. Uninstalling it from the main blog will delete the tables only if the plugin was installed globally.

This changeset is attempting to prevent full table removal when a plugin is uninstalled from any single blog on the network when the plugin is activated at the network level.

PR #1447, opened 2023-09-13
Check if entry/lead ID is empty in "gform_post_note_added" action

### LINKED ISSUE #1446: PHP Warnings/Notices on Gravity Forms note added
# Bug Report

## Expected Behavior

Upon using the "Save Progress" form feature in Gravity Forms along with an email Add-On provider there should be no warnings/issues thrown.

## Actual Behavior

A warning/notice is thrown because the email Add-On provider tries to create an entry note on a partial entry (which have no proper "entry_id" as they're saved in a separate "gf_draft_submissions" table), so the note-related actions/filters are still run by the email Add-On but provide said hooks with "null" instead of a proper "entry/lead id", and since there's no check in the Stream connector class for if the entry/lead id is a valid int it leads to the "$lead" on line 675 being set to "false" and then treated subsequently as an array:

```
[12-Sep-2023 23:58:38 UTC] PHP Warning:  Trying to access array offset on value of type bool in /../public_html/wp-content/plugins/stream/connectors/class-connector-gravityforms.php on line 676
[12-Sep-2023 23:58:38 UTC] PHP Warning:  Trying to access array offset on value of type null in /../public_html/wp-content/plugins/stream/connectors/class-connector-gravityforms.php on line 684
[12-Sep-2023 23:58:38 UTC] PHP Warning:  Trying to access array offset on value of type null in /../public_html/wp-content/plugins/stream/connectors/class-connector-gravityforms.php on line 685
```

File/location of issue:

https://github.com/xwp/stream/blob/4fb568dcc82bd8a7102e74c2a60fa5cd9493216c/connectors/class-connector-gravityforms.php#L657-L691

The fix should be as simple as checking if "$lead_id" in the above method is empty and returning early if that's the case. A further check on the whether the "$lead" variable is a valid array will also be needed as "\GFFormsModel::get_lead" can return a boolean if no entry is found for whatever reason. 

I'll open a PR regarding this issue shortly

1. Set PHP error_reporting to include notices & warnings
2. Install Gravity Forms, SendGrid Add-On, and Partial Entries Add-On
3. Setup small test form, in the Form Settings enable "Enable Save and Continue", and enable the "Partial Entries" feature for it as well
4. Set the "Save & continue notification email" to use the Add-On email providers instead of the built-in "WordPress" one
5. Visit form and attempt to use the "Save Progress" feature
6. Check confirmation message or PHP debug log for above Warnings or Notices

## Screenshots

Just a quick thing to point out is that when using the default "WordPress" email provider service for this Save & continue notification email it works as it should & no warning/notice is thrown, and that's because GF calls this internal method which explicitly checks for whether the "$entry_id" is empty or not before it tries to create a note, and most email Add-On providers seem to skip this check before trying to create a note.

In GF file "forms_model.php":

2023-09-13_10-10-l9tSRe1g

## System Information

- Stream plugin version: 3.9.3
- WordPress version: 6.3.1
- PHP version: 8.1.23
- Browser: Google Chrome | 117.0.5938.62
- Computer operating system: macOS Version 13.5.2 (Build 22G91)

zach-adams:
Fixes #1446.

Adds a simple "empty" check on the incoming $lead_id to prevent warnings/notices when "null" is passed under certain circumstances (detailed in issue).

**Note:** After reviewing the rest of the class I think there's a lot of safety/sanity checks & general updates that could/should be made to this Gravity Forms connector class and there's a lack of PHPUnit tests for this connector as well. Since this PR is relatively simple & straightforward I think it's fine to merge for the time being to fix the immediate notices/warnings problem, and I can submit a more detailed improvement issue/PR in the near future to cover a more proper general refactoring. Let me know if that sounds alright or if you'd prefer to wait on this issue/update until that more proper refactoring is ready

# Checklist

### DIFF
--- a/connectors/class-connector-gravityforms.php
+++ b/connectors/class-connector-gravityforms.php
@@ -672,6 +672,11 @@ class Connector_GravityForms extends Connector {
 		unset( $note );
 		unset( $note_type );
 
+		// Skip if no entry/lead id (e.g. Save and Continue notifications)
+		if( empty( $lead_id ) ) {
+			return;
+		}
+
 		$lead = \GFFormsModel::get_lead( $lead_id );
 		$form = $this->get_form( $lead['form_id'] );
 
--- a/connectors/class-connector-gravityforms.php
+++ b/connectors/class-connector-gravityforms.php
@@ -673,7 +673,7 @@ class Connector_GravityForms extends Connector {
 		unset( $note_type );
 
 		// Skip if no entry/lead id (e.g. Save and Continue notifications)
-		if( empty( $lead_id ) ) {
+		if ( empty( $lead_id ) ) {
 			return;
 		}
 


kasparsd (commented):
This looks good, thanks for your contribution!

kasparsd (approved):
Thanks, this looks great!

PR #1459, opened 2023-10-09
Improve client IP resolution for Stream logs

### LINKED ISSUE #1456: Improve event source IP reporting
# Feature Request

Determining user's IP reliably is difficult when the web server is behind layers of proxies as the request is made by the intermediary on behalf of the client. Here is how various services pass the origin IP to the server:

- Cloudflare https://developers.cloudflare.com/support/troubleshooting/restoring-visitor-ips/restoring-original-visitor-ips/
- WordPress VIP https://docs.wpvip.com/how-tos/configure-a-reverse-proxy/
- AWS CloudFront https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/example-function-add-true-client-ip-header.html

There is no reliable way for the WP plugin or WP core to detect the client IP so it would be useful to offer a filter for the current logic here:

https://github.com/xwp/stream/blob/bfc0530d617007040ce938f9eb20e7efb87c9fde/classes/class-log.php#L45-L49

kasparsd:
Fixes #1456.

- Rely on `$_SERVER['REMOTE_ADDR']` as the canonical source for the request origin IP.
- Show a notice in admin if `$_SERVER['REMOTE_ADDR']` is not specified.
- ~Fallback to `$_SERVER['HTTP_X_FORWARDED_FOR']` only if `$_SERVER['REMOTE_ADDR']` not present.~

# Checklist

## Release Changelog

- Fix: Prefer `$_SERVER['REMOTE_ADDR']` for the source of user IP and fallback to `$_SERVER['HTTP_X_FORWARDED_FOR']` only in case `REMOTE_ADDR` is not available. Introduce a `wp_stream_client_ip_address` filter to adjust the trusted client IP.

### DIFF
--- a/classes/class-admin.php
+++ b/classes/class-admin.php
@@ -220,6 +220,12 @@ class Admin {
 		$this->network     = new Network( $this->plugin );
 		$this->live_update = new Live_Update( $this->plugin );
 		$this->export      = new Export( $this->plugin );
+
+		// Check if the host has configured the `REMOTE_ADDR` correctly.
+		$client_ip = $this->plugin->get_client_ip_address();
+		if ( empty( $client_ip ) ) {
+			$this->notice( esc_html__( 'Stream can\'t determine a reliable client IP address! Please update the hosting environment to set the REMOTE_ADDR in $_SERVER variable or use the `wp_stream_client_ip_address` filter to specify the verified client IP address!', 'stream' ) );
+		}
 	}
 
 	/**
--- a/classes/class-admin.php
+++ b/classes/class-admin.php
@@ -547,9 +547,10 @@ class Admin {
 			return true;
 		}
 
-		$screen = get_current_screen();
-		if ( Alerts::POST_TYPE === $screen->post_type ) {
-			return true;
+		if ( is_admin() && function_exists( 'get_current_screen' ) ) {
+			$screen = get_current_screen();
+
+			return ( Alerts::POST_TYPE === $screen->post_type );
 		}
 
 		return false;
--- a/classes/class-admin.php
+++ b/classes/class-admin.php
@@ -224,7 +224,7 @@ class Admin {
 		// Check if the host has configured the `REMOTE_ADDR` correctly.
 		$client_ip = $this->plugin->get_client_ip_address();
 		if ( ! empty( $client_ip ) && $this->is_stream_screen() ) {
-			$this->notice( __( 'Stream plugin can\'t determine a reliable client IP address! Please update the hosting environment to set the <code>$_SERVER[\'REMOTE_ADDR\']</code> variable or use the <code>wp_stream_client_ip_address</code> filter to specify the verified client IP address!', 'stream' ) );
+			$this->notice( __( 'Stream plugin can\'t determine a reliable client IP address! Please update the hosting environment to set the $_SERVER[\'REMOTE_ADDR\'] variable or use the wp_stream_client_ip_address filter to specify the verified client IP address!', 'stream' ) );
 		}
 	}
 
--- a/classes/class-admin.php
+++ b/classes/class-admin.php
@@ -223,7 +223,7 @@ class Admin {
 
 		// Check if the host has configured the `REMOTE_ADDR` correctly.
 		$client_ip = $this->plugin->get_client_ip_address();
-		if ( ! empty( $client_ip ) && $this->is_stream_screen() ) {
+		if ( empty( $client_ip ) && $this->is_stream_screen() ) {
 			$this->notice( __( 'Stream plugin can\'t determine a reliable client IP address! Please update the hosting environment to set the $_SERVER[\'REMOTE_ADDR\'] variable or use the wp_stream_client_ip_address filter to specify the verified client IP address!', 'stream' ) );
 		}
 	}
--- a/classes/class-admin.php
+++ b/classes/class-admin.php
@@ -223,8 +223,8 @@ class Admin {
 
 		// Check if the host has configured the `REMOTE_ADDR` correctly.
 		$client_ip = $this->plugin->get_client_ip_address();
-		if ( empty( $client_ip ) ) {
-			$this->notice( esc_html__( 'Stream can\'t determine a reliable client IP address! Please update the hosting environment to set the REMOTE_ADDR in $_SERVER variable or use the `wp_stream_client_ip_address` filter to specify the verified client IP address!', 'stream' ) );
+		if ( ! empty( $client_ip ) && $this->is_stream_screen() ) {
+			$this->notice( __( 'Stream plugin can\'t determine a reliable client IP address! Please update the hosting environment to set the <code>$_SERVER[\'REMOTE_ADDR\']</code> variable or use the <code>wp_stream_client_ip_address</code> filter to specify the verified client IP address!', 'stream' ) );
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


classes/class-admin.php
```diff
@@ -536,9 +542,10 @@ public function is_stream_screen() {
 			return true;
 		}
 
-		$screen = get_current_screen();
-		if ( is_admin() && Alerts::POST_TYPE === $screen->post_type ) {
-			return true;
+		if ( is_admin() && function_exists( 'get_current_screen' ) ) {
+			$screen = get_current_screen();
```
kasparsd, line 546: Account for calls to this method when `get_current_screen()` is not available.

classes/class-log.php
```diff
@@ -87,9 +73,16 @@ public function log( $connector, $message, $args, $object_id, $context, $action,
 			return false;
 		}
 
+		$ip_address = $this->plugin->get_client_ip_address();
+
+		// Fallback to unsafe IP extracted from the request HTTP headers.
+		if ( empty( $ip_address ) ) {
+			$ip_address = $this->plugin->get_unsafe_client_ip_address();
```
kasparsd, line 80: Keep the previous behaviour for now. Consider removing this fallback in the future releases (as mentioned in the changelog).

classes/class-plugin.php
```diff
@@ -138,6 +145,9 @@ public function __construct() {
 		// Load logger class.
 		$this->log = apply_filters( 'wp_stream_log_handler', new Log( $this ) );
 
+		// Set the IP address for the current request.
+		$this->client_ip_address = wp_stream_filter_input( INPUT_SERVER, 'REMOTE_ADDR', FILTER_VALIDATE_IP );
```
kasparsd, line 149: Make the main plugin class responsible for knowing properties of the current request. Expose the knowledge via a helper method.

classes/class-log.php
```diff
@@ -140,7 +133,7 @@ function ( $var ) {
 			'connector' => (string) $connector,
 			'context'   => (string) $context,
 			'action'    => (string) $action,
-			'ip'        => (string) $this->ip_address,
+			'ip'        => (string) $ip_address,
```
schlessera, line 136: Both the boolean `false` as well as the `null` will be converted into an empty string here. Is that the intended behavior?

classes/class-log.php
```diff
@@ -87,9 +73,16 @@ public function log( $connector, $message, $args, $object_id, $context, $action,
 			return false;
 		}
 
+		$ip_address = $this->plugin->get_client_ip_address();
+
+		// Fallback to unsafe IP extracted from the request HTTP headers.
+		if ( empty( $ip_address ) ) {
```
schlessera, line 79: The `empty( $ip_address )` will be true for both the case where no IP address was provided as well for when an invalid IP address is provided. Is that the intended behavior?

classes/class-log.php
schlessera: I was wondering why you opted to do two separate calls in here, as opposed to having the fallback be part of `get_client_ip_address()` instead.

I thought the latter would have been easier for removing the fallback behavior at a later date. Also, it would allow `get_unsafe_client_ip_address()` to be private instead so we don't create an unwanted public interface for this.

classes/class-plugin.php
```diff
@@ -315,4 +325,48 @@ public function is_mustuse() {
...
+	}
+
+	/**
+	 * Get the client IP address from the HTTP request headers.
+	 *
+	 * There is no guarantee that this is the real IP address of the client.
+	 *
+	 * @return string|null
+	 */
+	public function get_unsafe_client_ip_address() {
```
schlessera, line 345: As mentioned elsewhere, if this fallback behavior would be called immediately while initially setting `$this->client_ip_address`, we wouldn't need to make it `public` and not have it part of the public interface.

schlessera (changes requested):

classes/class-log.php
kasparsd: The reason is we want to check just the "reliable" IP address in some instances with no fallbacks (when showing the admin notice, for example). This also clearly identifies the places where the use of `get_unsafe_client_ip_address()` should be wrapped into a conditional eventually.

readme.txt
```diff
@@ -65,12 +65,37 @@ With Stream’s powerful logging, you’ll have the valuable information you nee
...
+
+Update your server configuration to set the `$_SERVER['REMOTE_ADDR']` variable to the verified client IP address or use the `wp_stream_client_ip_address` filter to do that:
+
+`add_filter(
+    'wp_stream_client_ip_address',
+    function( $client_ip ) {
+        // Trust the X-Forwarded-For header.
+        if ( ! empty( $_SERVER['HTTP_X_FORWARDED_FOR'] ) ) {
+            return $_SERVER['HTTP_X_FORWARDED_FOR'];
+        }
```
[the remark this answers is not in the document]
kasparsd: Thanks for noticing this! I've updated the example to account for the CSV format of the IPs. For this example we should encourage users to rely on their own logic or helpers as necessary.

classes/class-admin.php
```diff
@@ -220,6 +220,12 @@ public function init() {
 		$this->network     = new Network( $this->plugin );
 		$this->live_update = new Live_Update( $this->plugin );
 		$this->export      = new Export( $this->plugin );
+
+		// Check if the host has configured the `REMOTE_ADDR` correctly.
+		$client_ip = $this->plugin->get_client_ip_address();
+		if ( ! empty( $client_ip ) && $this->is_stream_screen() ) {
+			$this->notice( __( 'Stream plugin can\'t determine a reliable client IP address! Please update the hosting environment to set the <code>$_SERVER[\'REMOTE_ADDR\']</code> variable or use the <code>wp_stream_client_ip_address</code> filter to specify the verified client IP address!', 'stream' ) );
```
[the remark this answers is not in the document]
kasparsd: I've removed any code formatting for simplicity.

classes/class-log.php
kasparsd: It is a bit tricky here -- we're attempting to preserve the `HTTP_X_FORWARDED_FOR` usage from before but now it will only happen if `REMOTE_ADDR` is not configured which might not be the case making this an equally breaking change.

I'm thinking we should probably do a breaking change release and make it consistent immediately to avoid confusion.

kasparsd:
@schlessera I feel like attempting the fallback behaviour might not be in our best interest here because it would actually be different from the previous behaviour. Let's make it a breaking change instead (does this warrant a 4.0.0 release?) and encourage everyone to use the newly introduced filter for adjusting the client IP address on environments where either `REMOTE_ADDR` is reporting wrong.

calvinalkan:
> @schlessera I feel like attempting the fallback behaviour might not be in our best interest here because it would actually be different from the previous behaviour. Let's make it a breaking change instead (does this warrant a 4.0.0 release?) and encourage everyone to use the newly introduced filter for adjusting the client IP address on environments where either `REMOTE_ADDR` is reporting wrong.

My 2C.

1. The fallback behaviour is insecure. People wil shoot themselves in the foot.
2. If PHP can't trust it's server to set REMOTE_ADDR correctly, or at all, all hope is lost. This should not be a consideration.
3. Yes, breaking change, 4.0.0

schlessera (approved):

PR #1462, opened 2023-10-11
Display metadata in the table

### LINKED ISSUE #947: Meta not included into Record object?
Hey guys,

I was working on a project that required logging some data and then export them to CSV, so I thought about extending Stream since it already had the ability to "log stuff" and export them to a csv plus we were already using stream to log changes to posts etc..

The custom connector works fine and saves all the data that I need and saves all the meta that I need. I was also about to extend the columns in the admin table to show the custom meta and I've noticed that metas aren't being retrieved through the Record object? Is there a reason why? 

I was hoping metas would be included into the record object so that I could include them into the CSV file but it's not like that. In the end I had to build a custom exporter too and query the database separately to include the metas.

akirk:
Fixes #947.

The metadata that can be stored along side records is burried in the `stream_meta` table. This make the data accessible in the table.

metadata-display
Screenshot 2026-05-24 at 04 34 29

# Checklist

I have run the code. I am unable to run the local development environment on a recent Macbook Pro. Maybe the environment should be updated in a separate PR.

> Error: Cask virtualbox depends on hardware architecture being one of [{:type=>:intel, :bits=>64}], but you are running {:type=>:arm, :bits=>64}.

Thus, I have also not been able to check the unit tests yet.

## Release Changelog

- New: Expose stream metadata in the admin UI.

## Release Checklist

### DIFF
--- a/.github/workflows/deploy-to-stream-dist.yml
+++ b/.github/workflows/deploy-to-stream-dist.yml
@@ -64,9 +64,20 @@ jobs:
           WORKING_BRANCH="$(git rev-parse --abbrev-ref HEAD)"
           SRC_DIR="$ROOT_DIR/source"
           DIST_DIR="$ROOT_DIR/dist"
-          DIST_BRANCH="${GITHUB_REF#refs/heads/}"
-          DIST_TAG="${GITHUB_REF#refs/tags/}"
           COMMIT_MESSAGE="$(git log -1 --oneline)"
+
+          if [ "${{ github.event_name }}" = "release" ]; then
+            # On a release event, GITHUB_REF is refs/tags/<tag>. Push the dist
+            # commit to the branch the release targets (e.g. master) and create
+            # the tag on top of it.
+            DIST_BRANCH="${{ github.event.release.target_commitish }}"
+            DIST_TAG="${GITHUB_REF#refs/tags/}"
+          else
+            # On a push event, GITHUB_REF is refs/heads/<branch>. Mirror the
+            # branch to dist without tagging.
+            DIST_BRANCH="${GITHUB_REF#refs/heads/}"
+            DIST_TAG=""
+          fi
 
           echo "ROOT_DIR=$ROOT_DIR" >> $GITHUB_ENV
           echo "WORKING_BRANCH=$WORKING_BRANCH" >> $GITHUB_ENV
@@ -121,7 +132,7 @@ jobs:
           if [ "${{ github.event_name }}" = "release" ]; then
             echo "Tagging a release: $DIST_TAG"
             git tag --force "$DIST_TAG"
-            git push --force --tags origin "$DIST_BRANCH"
+            git push --force origin "$DIST_BRANCH" "refs/tags/$DIST_TAG"
           else
             git push --force origin "$DIST_BRANCH"
           fi
--- a/abilities/class-ability-get-record.php
+++ b/abilities/class-ability-get-record.php
@@ -86,7 +86,7 @@ class Ability_Get_Record extends Ability {
 				'ip'        => array( 'type' => array( 'string', 'null' ) ),
 				'meta'      => array(
 					'type'        => 'object',
-					'description' => 'Per-record metadata key/value pairs.',
+					'description' => 'Per-record metadata key/value pairs. Single stored values are returned as scalars; legacy grouped values can be indexed arrays, while newer grouped values preserve sub-keys as objects.',
 				),
 			),
 		);
--- a/changelog.md
+++ b/changelog.md
@@ -2,16 +2,35 @@
 
 ## Unreleased
 
+### Enhancements
+
+- Add an opt-in **Display Metadata** setting for Stream records. New records preserve one-level grouped metadata sub-keys in the display/API shape; older records can still return legacy grouped values as indexed arrays, so API and MCP clients should handle both shapes.
+
+## 4.2.0 - May 28, 2026
+
+### New Features
+
+- Expose Stream abilities via the WordPress MCP Adapter when present, enabling AI tools to query Stream records through the Abilities API (XWPENG-13, [#1859](https://github.com/xwp/stream/pull/1859)).
+
 ### Bug Fixes
 
-- Fix unbounded growth of `stream` / `stream_meta` tables: the TTL-based auto-purge now runs via Action Scheduler with batched deletion (default 250,000 rows per batch via the existing `wp_stream_batch_size` filter), resolving database bloat on sites where the previous WP-Cron-driven purge silently failed or timed out on large tables (XWPENG-28).
+- Fix unbounded growth of `stream` / `stream_meta` tables: the TTL-based auto-purge now runs via Action Scheduler with batched deletion (default 250,000 rows per batch via the existing `wp_stream_batch_size` filter), resolving database bloat on sites where the previous WP-Cron-driven purge silently failed or timed out on large tables (XWPENG-28, [#1882](https://github.com/xwp/stream/pull/1882)).
 - Fix orphan `stream_meta` rows accumulating across repeated purge cycles: a terminal orphan reaper now runs at the end of every auto-purge chain, healing installs that have residual orphans from historical interrupted purges.
+- Skip Action Scheduler "is running?" probes on front-end pageloads: the `delete_all_records` and `clean_orphan_meta` settings fields no longer issue Action Scheduler queries on every front-end request, eliminating 3-4 unnecessary queries per pageview on every Stream-active site ([#1884](https://github.com/xwp/stream/issues/1884), [#1885](https://github.com/xwp/stream/pull/1885)).
 
 ### Enhancements
 
 - Add **Clean Orphaned Meta** link under **Settings → Advanced** for one-shot cleanup on already-bloated installs. The link is hidden while the auto-purge chain is running and reappears once the chain drains.
 - Replace the legacy `wp_stream_auto_purge` WP-Cron event with a recurring Action Scheduler action. Run history and failures are now visible under **Tools → Scheduled Actions**.
 - Auto-purge consults the existing `wp_stream_is_large_records_table` filter (default threshold: >1M rows) so small tables get a single inline DELETE while bloated tables go through the batched chain — same knob as the manual reset path.
+
+### Development
+
+- Add HTTPS to the local dev environment via mkcert and an Apache SSL vhost; force HTTPS across local dev and tests.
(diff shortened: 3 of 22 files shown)


akirk:
@kasparsd any chance of getting this merged? I've rebased this on the current release, it still works.

marcinkrzeminski:
@akirk thanks for the PR. I agree that it would be best to make this feature optional. Do you have the capacity to add a setting that enables it?

PatelUtkarsh:
Hi @akirk, are you still interested in landing this? The metadata display feature is useful and your recent rebase suggests you might be.

To move forward, two things:

1. Rebase onto current `develop`; the branch is conflicting.
2. Review the AI-assisted code review below (I reviewed all comments) and respond to the findings. Push back on anything that doesn't hold up.

Happy to help with either or take it over if you'd prefer to hand it off.

---

## AI-assisted code review

Findings on the diff between this PR's HEAD and current `develop`. Each was double-checked against the actual code on this branch and the WordPress meta API docs before posting. Classified as blocker, warning, or nit. Numbered for easy reference (e.g. "comment 3").

### Blockers

**1. `classes/class-db-driver-wpdb.php` (insert_record meta loop) — UTF-8 mid-byte truncation via `substr()`**

`substr( $val, 0, 200 )` cuts on byte boundaries. Any multibyte character that straddles byte 200 produces invalid UTF-8 in `stream_meta.meta_value`. The schema is `varchar(200)` so MySQL truncates at the character boundary anyway; doing the chop in PHP is both redundant and unsafe. Either use `mb_strimwidth( $val, 0, 200, '' )` / `mb_substr`, or drop the PHP-side truncation entirely.

**2. `classes/class-settings.php` and `classes/class-list-table.php` — bypasses network-aware accessor**

`List_Table::should_display_metadata()` reads `$this->plugin->settings->options['advanced_display_metadata']`. `Settings::$options` is populated from `get_site_option( 'wp_stream' )` only inside `is_network_admin()`; in the regular per-site records page it reads `get_option( 'wp_stream' )` — the per-site option store. On a network-activated install the authoritative store is `wp_stream_network`, so the toggle won't take effect.

Use `$this->plugin->settings->get_setting_value( 'advanced_display_metadata', 0 )` instead. That accessor was introduced for the Abilities/MCP work and resolves the network-vs-site read centrally.

**3. `classes/class-record.php` (get_meta) — `$single=false` wraps the result in an extra array layer**

```php
public function get_meta( $meta_key = '', $single = false ) {
    $meta = get_metadata( 'record', $this->ID, $meta_key, $single );
    if ( '' === $meta_key || ! empty( $meta ) ) {
        return '' === $meta_key ? self::normalize_meta( $meta ) : $meta;
    }
    $all_meta = self::normalize_meta( get_metadata( 'record', $this->ID ) );
    if ( ! array_key_exists( $meta_key, $all_meta ) ) {
        return $single ? '' : array();
    }
    return $single ? $all_meta[ $meta_key ] : array( $all_meta[ $meta_key ] );
}
```

Trace `get_meta('user_meta', false)` against a record with bracketed sub-keys:

1. `get_metadata( ..., 'user_meta', false )` returns `[]` (no row with that literal key).
2. The `! empty( $meta )` short-circuit is false, so it falls through.
3. Second `get_metadata` returns the bracketed rows; `normalize_meta` produces `['user_meta' => ['agent' => 'wp_cli', 'display_name' => 'HAL']]`.
4. `$single` is false, returns `array( ['agent' => ..., 'display_name' => ...] )` — wrapped in an extra outer array.

WP's `get_metadata( ..., $single=false )` returns a flat array of values; the new behavior is `[ [keyed] ]`. The existing test (`test_query_fetches_array_meta_when_requested`) only exercises `$single=true`, which works. The `$single=false` path is the bug.

Also: the fall-through executes a second SQL query on every key miss. On a list-table render of 50 rows, that's an extra 50 queries when none of the records have the requested key.

Simplest fix: drop the magic. Keep `get_meta()` as a pass-through to `get_metadata()` and document that callers needing bracketed-key reconstruction call `Record::normalize_meta()` themselves.

### Warnings

**4. `classes/class-db-driver-wpdb.php` (insert_record meta loop) — `user_meta` shape splits across the write boundary**

Before this PR, nested-array meta was written with sub-keys discarded. `meta = ['user_meta' => ['user_email' => 'admin@example.com', 'display_name' => 'admin', ...]]` produced one row per value, all keyed `user_meta`, sub-keys lost at write time. Verified empirically against the live `stream_meta` table on a current install: rows keyed `user_meta` hold bare values like `admin@example.com`, `admin`, `Administrator`.

After this PR, the same input writes `user_meta[user_email]`, `user_meta[display_name]`, etc. Reads via `Record::normalize_meta()` reconstruct the nested object.

Both shapes coexist in `stream_meta` after deploy. Observable in two places:

- The new "Display Metadata" column in the list table renders both shapes in adjacent rows. An admin browsing across the upgrade boundary will see `"user_meta": ["admin@example.com", "admin", "admin", "Administrator"]` for old records and `"user_meta": {"user_email": "admin@example.com", ...}` for new ones.
- `Ability_Get_Record` (MCP/REST `stream/get-record`) returns the same split. The output schema declares `meta` as `object`, which both shapes satisfy at the top level, but `user_meta` itself changes from a JSON array to a JSON object.

Where this does not matter on current core: the list table's own Summary/User columns, connectors, exporters, and alerts do not read `user_meta` from the stored rows. They construct `Author` from `get_userdata()` live, so the storage difference is invisible to them. The data has been effectively unused at the admin-UI layer.

Recommendation: don't ship a migration; document the split. Old records are still useful (the values are searchable strings even without sub-keys), so a lossy-delete migration removes data unnecessarily. Position-based remapping is fragile (the old write loop's empty-value filter shifts positions per-record) and shouldn't be attempted. A changelog note plus a paragraph in the MCP ability docs is enough: "Records logged before X.Y.Z return `user_meta` as an indexed array; records from X.Y.Z onward as an object. MCP clients should handle both shapes."

**5. `classes/class-record.php` (get_by_id) — meta shape change is a public-API change for MCP/REST**

Before: `$row['meta'] = (array) get_metadata( 'record', $row['ID'] );`
After: `$row['meta'] = self::normalize_meta( get_metadata( 'record', $row['ID'] ) );`

`abilities/class-ability-get-record.php` (the `stream/get-record` ability) returns `$row['meta']` directly. After this PR, scalar single-row meta gets unwrapped from `['v']` to `'v'` (via `normalize_meta_value()`'s `count === 1` collapse), and bracketed rows get re-grouped. MCP/REST clients of `stream/get-record` see a different JSON shape after deploy.

The unwrap arguably matches what most callers expected anyway (`"space_helmet": "false"` reads more naturally than `"space_helmet": ["false"]`), so this is an intentional API improvement, not a regression. But it is a shape change. Mention it in the changelog alongside finding 4 and the MCP doc.

**6. `classes/class-record.php` (normalize_meta_value) — collapses 1-element arrays unconditionally**

```php
if ( is_array( $value ) ) {
    $value = array_map( 'maybe_unserialize', $value );
    if ( 1 === count( $value ) ) {
        return reset( $value );
    }
    return $value;
}
```

This is what produces the unwrap in finding 5. It diverges from `get_metadata`'s documented shape but matches what's intuitive. Decision needed: keep the collapse (and document) or drop it (and let callers index). My weak preference is keep + document; the collapsed shape is friendlier for MCP clients.

**7. `classes/class-query.php` (count query) — removed `{$join}`**

```diff
-		FROM $wpdb->stream
-		{$join}
-		WHERE 1=1 {$where}";
+		FROM $wpdb->stream
+		WHERE 1=1 {$where}";
```

`$join` is `''` in stock Stream so this is a no-op today. But the `wp_stream_db_query_where` / `wp_stream_db_count_query` filters and custom `DB_Driver` subclasses are documented extension points; a downstream consumer that injects a `JOIN` via filter expected count and select to apply symmetrically. The commit message says "Fix count query" — what was the actual count miscount you were addressing? If it was meta-join row multiplication, `SELECT COUNT(DISTINCT $wpdb->stream.ID)` is the structural fix and preserves the extension point.

**8. `classes/class-query.php` (meta fetch loop) — ordering + per-record cap**

The second SQL uses `ORDER BY meta_id ASC`. Fine, but two notes:

1. With `records_per_page = 50` on a chatty site (5-10 meta rows per record) this pulls ~500 rows per admin page render. Acceptable given the opt-in setting, but a single pathological record could spike the response size. Consider documenting the expectation or adding a soft cap.
2. Legacy multi-row records depend on `meta_id ASC` to produce a stable read order. The test fixture (`dummy_meta_data()`) returns a flat array, so the legacy path isn't exercised. Please add a test that inserts the legacy shape and asserts the read order.

**9. `classes/class-query.php` (synthetic `record_id` alias)**

```php
if ( $with_meta ) {
    if ( ! empty( $fields ) && ! in_array( 'ID', $fields, true ) ) {
        $selects[] = "$wpdb->stream.ID AS record_id";
    }
}
```

Cleaner: if `with_meta` is on, just append `ID` to `$selects` when missing. Eliminates the aliasing and the `unset( $item->record_id )` cleanup further down.

**10. `classes/class-record.php` (bracketed-key parser) — edge cases**

`'foo[bar][baz]'` (nested brackets) parses `main_key = 'foo'`, `sub_key = 'bar][baz'`. A meta key from another plugin that happens to contain `[...]` could also be misparsed. Recommend either rejecting nested brackets at the writer or documenting the format as one-level-only and adding a test that proves either decision.

### Nits

**11.** `connectors/class-connector-buddypress.php`: pure whitespace re-alignment. Split or revert; drift in feature PRs hurts bisects.

**12.** `classes/class-settings.php`: field `name` is `display_metadata` but the read uses `advanced_display_metadata`. Correct (Stream prefixes by tab slug), but a code comment near the read would prevent confusion.

**13.** `classes/class-query.php`: the new comma-string `fields` parsing isn't documented in the args docblock. Either remove (internal callers use the array form) or document.

**14.** `classes/class-record.php` (normalize_meta): if both a flat `user_meta` row and a bracketed `user_meta[agent]` row coexist for the same record, the flat value gets silently overwritten by the array. Document the precedence or guard at the writer.

**15.** Test coverage: please add cases for the legacy multi-row format (insert old shape via raw SQL, read via `get_by_id`), multibyte values >200 bytes, a key with literal `[` from an unrelated source, a record with both flat and bracketed keys, and `get_meta('user_meta', false)` against bracketed storage (to lock in the fix for finding 3).

akirk:
Yes! Thanks for that review. I didn't have enough time to test it thoroughly enough after my changes, so I hadn't continued the conversation yet.

PR #1465, opened 2023-10-24
Install Composer v2 on TravisCI

szepeviktor:
As Composer v2 was used to update the lock file
https://github.com/xwp/stream/blob/3273731a0182353ee78cec3a0f9f77956407a932/composer.lock#L7445


szepeviktor:
I thought this would fix unit tests...

PR #1466, opened 2023-11-03
Adjust exclude rules filtering to avoid passing null to strlen() to fix PHP 8 deprecation notice

### LINKED ISSUE #1349: PHP 8.1: PHP Deprecated:  strlen(): Passing null to parameter #1 ($string) of type string is deprecated
# Bug Report

## Expected Behavior

No error message.

## Actual Behavior

```
PHP Deprecated:  strlen(): Passing null to parameter #1 ($string) of type string is deprecated in /usr/local/wordpress/wp-content/plugins/stream/classes/class-log.php on line 215
```

1. Just logging in and traversing core admin pages.

## System Information

- Stream plugin version: 3.9.0
- WordPress version: 6.0
- PHP version: 8.1

ocean90:
Fixes #1349.

This fixes a deprecation notice in PHP 8:

>  PHP Deprecated:  strlen(): Passing null to parameter #1 ($string) of type string is deprecated in classes/class-log.php on line 217

The value of my exclude rule was:

```
Array
(
    [connector] =>
    [context] =>
    [action] =>
    [ip_address] => 127.0.0.1
    [author] =>
    [role] =>
)
```

The fix is to check for null first.

# Checklist

## Release Changelog

- Fix: Avoid deprecation notices in PHP 8.

## Release Checklist

### DIFF
--- a/classes/class-log.php
+++ b/classes/class-log.php
@@ -194,7 +194,12 @@ class Log {
 				'role'       => ( ! empty( $exclude_rule['author_or_role'] ) && ! is_numeric( $exclude_rule['author_or_role'] ) ) ? $exclude_rule['author_or_role'] : null,
 			);
 
-			$exclude_rules = array_filter( $exclude, 'strlen' );
+			$exclude_rules = array_filter(
+				$exclude,
+				static function ( $value ) {
+					return null === $value || strlen( (string) $value ) > 0;
+				}
+			);
 
 			if ( $this->record_matches_rules( $record, $exclude_rules ) ) {
 				$exclude_record = true;


classes/class-log.php
```diff
@@ -194,7 +194,12 @@ public function is_record_excluded( $connector, $context, $action, $user = null,
 				'role'       => ( ! empty( $exclude_rule['author_or_role'] ) && ! is_numeric( $exclude_rule['author_or_role'] ) ) ? $exclude_rule['author_or_role'] : null,
 			);
 
-			$exclude_rules = array_filter( $exclude, 'strlen' );
+			$exclude_rules = array_filter(
+				$exclude,
+				static function ( $value ) {
+					return null === $value || strlen( (string) $value ) > 0;
```
kasparsd, line 200: This logic now allows `null` values to be included in the filtered output:

```php
$list = [ 1, 'two', false, null, 0 ];

$rules = array_filter(
    $list,
    function ( $value ) {
        return null === $value || strlen( (string) $value ) > 0;
    }
);

var_dump( $rules );
```

returns:

```
array(4) {
  [0]=>
  int(1)
  [1]=>
  string(3) "two"
  [3]=>
  NULL
  [4]=>
  int(0)
}
```

while we would expect the `null` value to be removed.

Could we change this to `strlen( (string) $value ) > 0`?

kasparsd (changes requested):
Thanks for the pull request @ocean90!

It appears that the proposed change also adjusts the business logic -- could we update it to keep the previous logic, please?

classes/class-log.php
ocean90: I think you're reading the previous logic wrong. Previously null was also allowed which is why I added the specific check for null. Removing that would actually be a change in the logic.

kasparsd: @ocean90 I just checked this again -- previously it was using `array_filter( $exclude, 'strlen' )` which does _remove_ `null` values:

```php
$list = [ 1, 'two', false, null, 0 ];
var_dump( 'before', array_filter( $list, 'strlen' ) );
```

which returns:

```
Deprecated: strlen(): Passing null to parameter #1 ($string) of type string is deprecated in /Users/kaspars/Desktop/filter.php on line 12
string(6) "before"
array(3) {
  [0]=>
  int(1)
  [1]=>
  string(3) "two"
  [4]=>
  int(0)
}
```

while the updated version keeps the `null` in the list as mentioned above.

delawski:
I'm closing this one in favor of #1513.

PR #1312, opened 2022-03-09
Fixup fc606db

lkraav:
This reverts commit fc606dbd175b1ce533ea88198f8bf91a78fb1439.

We found a more correct location to improve a previously existing sanity check.

Fixes #1080, but better. Sorry @kasparsd :smile:

You were right, there was a connection between #1038 and #1080 all along. I now traced down how all meta is added to a record, and found we're already trying to prevent empty values. Unfortunately

* existing `$stream_meta is_null()` check along is weak
* `$user_meta` being a nested array, its actual contents isn't checked at all

`strlen` trick seems to be a solid one, considering hundreds of upvotes on PHP docs. I always disliked having to do a separate empty string check for `is_scalar()` result.

Now that previous version is in, this PR doesn't have urgency for its own release.

PR #1473, opened 2024-01-10
Refactored array_filter() to explicitly check !is_null() AND strlen() in class-log.php

### LINKED ISSUE #1349: PHP 8.1: PHP Deprecated:  strlen(): Passing null to parameter #1 ($string) of type string is deprecated

justinmaurerdotdev:
… because passing 'strlen' as a callable caused warnings on null values.

Fixes  #1349.

The previously-merged commits did not, in fact, fix this issue. Null values were being passed through array_filter() with 'strlen' as the callable, but `strlen()` no longer accepts null values. Should be fixed now, with an explicit null check added.

OLD:
```
$exclude_rules = array_filter($exclude, 'strlen' );
```

NEW:
```
$exclude_rules = array_filter( $exclude, function ( $val ) {
	return !is_null( $val ) && strlen( $val );
});
```

# Checklist

- [n/a] Project documentation has been updated to reflect the changes in this pull request, if applicable.
- [X] I have tested the changes in the local development environment (see `contributing.md`).

## Release Changelog

## Release Checklist

### DIFF
--- a/classes/class-log.php
+++ b/classes/class-log.php
@@ -194,8 +194,8 @@ class Log {
 				'role'       => ( ! empty( $exclude_rule['author_or_role'] ) && ! is_numeric( $exclude_rule['author_or_role'] ) ) ? $exclude_rule['author_or_role'] : null,
 			);
 
-			$exclude_rules = array_filter($exclude, function ($val) {
-				return !is_null($val) && strlen($val);
+			$exclude_rules = array_filter( $exclude, function ( $val ) {
+				return !is_null( $val ) && strlen( $val );
 			});
 
 			if ( $this->record_matches_rules( $record, $exclude_rules ) ) {
--- a/classes/class-log.php
+++ b/classes/class-log.php
@@ -194,7 +194,9 @@ class Log {
 				'role'       => ( ! empty( $exclude_rule['author_or_role'] ) && ! is_numeric( $exclude_rule['author_or_role'] ) ) ? $exclude_rule['author_or_role'] : null,
 			);
 
-			$exclude_rules = array_filter( $exclude, 'strlen' );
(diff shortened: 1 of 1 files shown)


lkraav:
Ouch, I wonder if `strlen()` changes also kill plans in #1312 :(

justinmaurerdotdev:
Hmm, yeah. If it is being used to filter null values, then it likely will cause the same problem. The two refactoring options, as far as I can tell are:
1. My above solution, or some similar variation where you check for null in the filter's callable, or
2. An additional, `array_filter()` preceding the existing one to filter null values 

I think Option 1 wins both for performance (Option 2 re-iterating the whole array) and presentation. There's no native `is_not_null` function. If there were, Option 2 would at least win on aesthetics:
```
$exclude_rules = array_filter($exclude, 'is_not_null' );
$exclude_rules = array_filter($exclude, 'strlen' );
```

In reality, it has to be something like 
```
$exclude_rules = array_filter( $exclude, function ( $val ) {
	return !is_null( $val );
});
$exclude_rules = array_filter($exclude, 'strlen' );
```

So, we might as well combine them and save the extra loop. Though, this may be a sign of some upstream issue with how the how/why the null values are there in the first place... I'm new to this codebase, so I can't speak to that at all.

ocean90:
I already created a PR for this in #1466 which also ensures that only strings are passed to `strlen()`.

PR #1475, opened 2024-01-12
CLI: Fix PHP Type error

### LINKED ISSUE #1474: CLI: PHP Uncaught TypeError
# Bug Report
 If I type `wp stream query` on the CLI, I get an PHP error.

`Fatal error: Uncaught TypeError: array_key_exists(): Argument #2 ($array) must be of type array, stdClass given in /wp-content/plugins/stream/classes/class-cli.php:137`

# Possible Fix
A record is an Object, not an Array, so we should use `property_exists()` instead `array_key_exists`.

Before
```PHP
foreach ( $fields as $field ) {
	if ( ! array_key_exists( $field, $record ) ) {
		$record->$field = null;
	}
}
```

After
```PHP
foreach ( $fields as $field ) {
	if ( ! property_exists( $record, $field ) ) {
		$record->$field = null;
	}
}
```

## System Information

- Stream plugin version: 4.0.0
- WordPress version: 6.4.2
- PHP version: 8.2
- WP-CLI: 2.9.0

Soean:
A record is an Object, not an Array, so we should use `property_exists()` instead `array_key_exists`.

Fixes #1474.

## Release Changelog

- Fix: PHP Uncaught TypeError in CLI command #1474

### DIFF
--- a/classes/class-cli.php
+++ b/classes/class-cli.php
@@ -134,7 +134,7 @@ class CLI extends \WP_CLI_Command {
 
 			// Catch any fields missing in records.
 			foreach ( $fields as $field ) {
-				if ( ! array_key_exists( $field, $record ) ) {
+				if ( ! property_exists( $record, $field ) ) {
 					$record->$field = null;
 				}
 			}


roborourke:
Hey @kasparsd, the CLI  `query` command is completely broken right now because of this. Will be great to get a patch release out asap if possible.

If you're open to adding some more maintainers I'd be happy to help in that regard from the Human Made side.

kasparsd (approved):
This is great, thanks!

PR #1496, opened 2024-07-16
Fix vscode launch.json and update Docker troubleshooting 

tharsheblows:
This fixes the launch.json file for vscode. 

It also adds instructions on how to pull fresh images in case you have old ones sitting around causing issues, eg if the version of Xdebug is incorrect when starting.

This is a PR to the fix/dev-env branch.


delawski (approved):
Looks great, thank you!

PR #1505, opened 2024-07-16
publish docker image on master only again

tharsheblows:
I think we should be more selective about when to publish a new Docker image; it feels to me that any changes should go through a round of code review before being published? 

It might be that it should be published on `develop` instead which allows us to update the local environment without a release but after going through the code review and testing process.

That having been said, pushing this branch, `fix/publish-on-master`, [also published the Docker image](https://github.com/xwp/stream/actions/runs/9962566359) so it might need something a bit different.

### DIFF
--- a/.github/workflows/docker-images.yml
+++ b/.github/workflows/docker-images.yml
@@ -36,5 +36,5 @@ jobs:
         run: docker compose --file docker-compose.build.yml build
 
       - name: Publish images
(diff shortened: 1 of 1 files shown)


delawski (approved):
Definitely! I'm sorry, I was supposed to uncomment that line as soon as the image build process is finished by a GitHub action but I didn't manage to do that today. Your fix is definitely what we want 👍

Issue #1278, opened 2021-08-31
Fix PHP unit tests with PHP 8


kasparsd:
# Feature Request

Make the phpunit tests work with PHP 8. It appears that the `php-token-stream` library referenced by `phpunit` isn't compatible with PHP 8:

```
Stack trace:
#0 /var/www/html/wp-content/plugins/stream-src/vendor/phpunit/php-token-stream/src/Token/Stream.php(118): PHP_Token_Stream->scan('<?php\n/**\n * Al...')
#1 /var/www/html/wp-content/plugins/stream-src/vendor/phpunit/php-code-coverage/src/CodeCoverage.php(764): PHP_Token_Stream->__construct('<?php\n/**\n * Al...')
#2 /var/www/html/wp-content/plugins/stream-src/vendor/phpunit/php-code-coverage/src/CodeCoverage.php(673): SebastianBergmann\CodeCoverage\CodeCoverage->getLinesToBeIgnored('/var/www/html/w...')
#3 /var/www/html/wp-content/plugins/stream-src/vendor/phpunit/php-code-coverage/src/CodeCoverage.php(321): SebastianBergmann\CodeCoverage\CodeCoverage->applyIgnoredLinesFilter(Array)
#4 /var/www/html/wp-content/plugins/stream-src/vendor/phpunit/php-code-coverage/src/CodeCoverage.php(1104): SebastianBergmann\CodeCoverage\CodeCoverage->append(Array, 'UNCOVERED_FILES...')
#5 /var/www/html/wp-content/plugins/stream-src/vendor/phpunit/php-code-coverage/src/CodeCoverage.php(256): SebastianBergmann\CodeCoverage\CodeCoverage->initializeData()
#6 /var/www/html/wp-content/plugins/stream-src/vendor/phpunit/phpunit/src/Framework/TestResult.php(650): SebastianBergmann\CodeCoverage\CodeCoverage->start(Object(WP_Stream\Test_Alert_Trigger_Action))
#7 /var/www/html/wp-content/plugins/stream-src/vendor/phpunit/phpunit/src/Framework/TestCase.php(868): PHPUnit_Framework_TestResult->run(Object(WP_Stream\Test_Alert_Trigger_Action))
#8 /var/www/html/wp-content/plugins/stream-src/vendor/phpunit/phpunit/src/Framework/TestSuite.php(733): PHPUnit_Framework_TestCase->run(Object(PHPUnit_Framework_TestResult))
[...]

PR #1510, opened 2024-07-18
Do not start Xdebug automatically, use sessionkey instead

delawski:
This change prevents console errors when running PHPUnit tests as described in https://github.com/xwp/stream/pull/1508#discussion_r1682716907

Xdebug will no longer be automatically enabled whenever a request comes in. Instead, a request should contain a sessionkey added by a browser extension like [Xdebug helper](https://chromewebstore.google.com/detail/xdebug-helper/eadndfjplgieldjbigjakmdgkmoaaaoc?hl=en).

# Checklist


tharsheblows (commented):
@delawski I use xdebug with phpunit a lot 😅 -- but I do think we should sort the underlying issue of "don't show warnings or deprecations before the unit tests" if possible. This isn't an issue with running the tests in php 7.4 but is in php 8.2 even with the correct phpunit.xml settings; they're still before the tests for some reason. 

Can we address this in this issue? https://github.com/xwp/stream/issues/1278

delawski:
This is superseded by #1511.

PR #1512, opened 2024-07-19
Require PHP ≥ 7.0 and WordPress ≥ 4.6

### LINKED ISSUE #1499: Deprecate PHP 5.6
Following WordPress Core, we should deprecate PHP 5.6 and instead use PHP 7.0 as the minimum version:

https://wordpress.org/about/requirements/

delawski:
Fixes #1499

This PR updates the minimum PHP version requirement to 7.0. I tested this locally and besides getting some PHP notices from non-plugin related code everything seemed to be working as expected.

I also tried downgrading WordPress core to 4.5, which is the version we required so far, but it failed. WP admin did not load and I got dozens of issues (I tried on PHP 7.1) Maybe it's something about my setup so I'd appreciate someone else double-checking that as well.

When I upgraded to 4.6, everything went back to normal. That's why, at least for now, I've bumped the minimal requirement to 4.6 for WordPress core.

I also did some quick smoke test of WordPress 6.6 and did not notice any issues. That's why I increased the "Tested up to" label to 6.6.

# Checklist

## Release Changelog

- Update: Set minimum PHP version requirement to 7.0 and WordPress core version to 4.6.
- Update: Test the plugin on WordPress 6.6 and update the "tested up to" label accordingly.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     }
   ],
   "require": {
-    "php": ">=5.6.20",
+    "php": ">=7.0",
     "composer/installers": "~1.0",
     "ext-json": "*"
   },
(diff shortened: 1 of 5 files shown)


tharsheblows (approved):
Works perfectly!

Re this:

> I also tried downgrading WordPress core to 4.5, which is the version we required so far, but it failed. WP admin did not load and I got dozens of issues (I tried on PHP 7.1) Maybe it's something about my setup so I'd appreciate someone else double-checking that as well.

I think if it failed for you for whatever reason, it's best to bump the minimum version to what works.

PR #1518, opened 2024-07-24
Document Connectors

### LINKED ISSUE #1515: Document current Connectors
# Feature Request

Create a document with the current Connectors and hooks.

tharsheblows:
Fixes #1515 

Adds a cheap and cheerful script that runs locally to create the Connectors documentation and also adds the documentation itself.

# Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "switch-to:php8.2": "docker compose build --build-arg PHP_VERSION=8.2 --build-arg XDEBUG_VERSION=3.3.2 && npm run start",
     "switch-to:php7.4": "docker compose build && npm run start",
     "which-php": "npm run cli -- php --version",
+    "document:connectors": "php ./local/scripts/generate-connector-docs.php",
     "stop": "docker compose down",
     "stop-all": "docker stop $(docker ps -a -q)",
     "logs": "docker compose logs --follow",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -57,6 +57,7 @@
     "switch-to:php8.2": "docker compose build --build-arg PHP_VERSION=8.2 --build-arg XDEBUG_VERSION=3.3.2 && npm run start",
     "switch-to:php7.4": "docker compose build && npm run start",
     "which-php": "npm run cli -- php --version",
+    "document:connectors": "php ./local/scripts/generate-connector-docs.php",
```
delawski, line 60: **💡 thought:** I was wondering if it'd be worth running this script from within the Docker container. Just so that we do not rely on the developer's local environment setup. Also, I guess by running it through Docker, we could make this job part of a GitHub action in future.

delawski (approved):
This is fantastic, @tharsheblows! 🤩 I left a thought but it is ready to be merged as is 👍

package.json
tharsheblows: @delawski It's not copied to the container so we'd need to pipe it in and I can only figure out how to do it via `docker exec ... ` which requires the container name or id which can change a bit (eg mine is `stream-wordpress-1` because I already had a `stream-wordpress` I guess.

Or we could copy it into the container but ... I don't know that seems a bit much?

PR #1521, opened 2024-07-25
Use GitHub action for doing WP.org deployment

### LINKED ISSUE #1514: Modernize SVN deployment workflow
# Feature Request

So far, the plugin was deployed to wordpress.org SVN repository using the [`grunt-wp-deploy`](https://github.com/stephenharris/grunt-wp-deploy) task. It was manually run on Committer's local machine to deploy to wp.org.

I don't think we should keep the Grunt-based deployment workflow in place. The task is old and has not been updated in the last few years.

I'd suggest using the [`wordpress-plugin-deploy`](https://github.com/marketplace/actions/wordpress-plugin-deploy) GitHub Action instead of the rather old and not maintained `grunt-wp-deploy` task. This would allow us make the release process more transparent (since everyone has access to the GH action log) and more streamlined (it would be run whenever a new release tag is pushed to the repo by an authorized user).

delawski:
Fixes #1514.

This PR adds the [wordpress-plugin-deploy](https://github.com/marketplace/actions/wordpress-plugin-deploy) GitHub action which is supposed to replace the legacy `grunt-wp-deploy`-based workflow.

# Checklist

## Release Changelog

- New: Use the `wordpress-plugin-deploy` GitHub action for deployments to WP.org
- Development: Update Node version to `20`


Gruntfile.js
```diff
@@ -50,80 +50,19 @@ module.exports = function( grunt ) {
...
-	grunt.loadNpmTasks( 'grunt-contrib-compress' );
 	grunt.loadNpmTasks( 'grunt-contrib-uglify' );
-	grunt.loadNpmTasks( 'grunt-wp-deploy' );
 
 	// Register tasks
 	grunt.registerTask( 'default', [ 'clean', 'uglify', 'cssmin' ] );
-	grunt.registerTask( 'build', [ 'default', 'copy' ] );
-	grunt.registerTask( 'release', [ 'build', 'compress' ] );
-	grunt.registerTask( 'deploy', [ 'build', 'wp_deploy', 'clean' ] );
 };
```
tharsheblows, line 68: 💖💖💖 All the red in this file!

tharsheblows (approved):
@delawski This is great! I'm not merging for the moment as I kind of think we should update the [docker build action](https://github.com/xwp/stream/blob/update/1514-build-workflow/.github/workflows/docker-images.yml) to only run when that workflow is updated rather than any workflow? What do you think?

Gruntfile.js
delawski: Hopefully in `4.1.0` the whole `Gruntfile.js` will be in red 😄

delawski:
> I'm not merging for the moment as I kind of think we should update the [docker build action](https://github.com/xwp/stream/blob/update/1514-build-workflow/.github/workflows/docker-images.yml) to only run when that workflow is updated rather than any workflow? What do you think?

Great suggestion! I did that in c6bc437.

PR #1530, opened 2024-07-30
Remove duplicate logs when saving posts

### LINKED ISSUE #1525: Publishing/Updating in Block Editor Creates Two Records
# Bug Report

When creating a new post in the block editor, 2 log entries are created:
 
Screenshot 2024-07-29 at 13 12 45

Updating an existing post in the block editor (e.g. a title change) generates 2 log entries as well:

Screenshot 2024-07-29 at 13 14 31

## Expected Behavior

The same (single) action performed in the block editor should generate only one log entry. This should be the same as for e.g. updating a post using Quick Edit feature, see:

Screenshot 2024-07-29 at 13 16 13

## Actual Behavior

The same (single) action performed in the block editor creates 2 log entries because 2 REST requests are being sent by the block editor whenever the "Save" button is clicked.

1. Create a new post using the block editor
2. Publish the post
3. In the Stream log, 2 entries are being presented instead of 1
4. Go to the same post again
5. Change its title and save
6. In the Stream log, 2 entries are being displayed instead of 1

## System Information

- Stream plugin version: 4.0.1 RC1
- WordPress version: 6.6
- PHP version: 8.2.21
- Browser: Chrome 126
- Computer operating system: MacOS Sonoma

tharsheblows:
Fixes #1525 

The block editor makes two requests each time a post is saved: once to save the post and once to save the legacy metaboxes. This PR checks for the `meta-box-loader` query variable and does not add a log if that is present.

This also adds the Classic Editor plugin for local testing.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,7 @@
     "wp-coding-standards/wpcs": "^3.1",
     "wp-phpunit/wp-phpunit": "^6.6",
     "wpackagist-plugin/advanced-custom-fields": "6.3.4",
+    "wpackagist-plugin/classic-editor": "1.6.4",
     "wpackagist-plugin/easy-digital-downloads": "3.3.1",
     "wpackagist-plugin/jetpack": "13.6",
     "wpackagist-plugin/user-switching": "1.8.0",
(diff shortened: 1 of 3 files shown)


delawski (approved):
This is great, thank you!

Issue #1503, opened 2024-07-16
Large tables not able to be reset


tharsheblows:
# Bug Report

https://wordpress.org/support/topic/how-to-manually-delete-records/
https://wordpress.org/support/topic/database-bloat-unable-to-archive-records/

## Expected Behavior

Records should be deleted when 'Reset Stream Database' is clicked. Records older than the saved records time should be deleted.

## Actual Behavior

from https://wordpress.org/support/topic/how-to-manually-delete-records/
> I currently have 2,453,623 items in the stream records. I cannot delete these using the ‘Reset Stream Database’ option in the Advanced tab as the request is timing out. Is there a manual way to delete these records safely in the DB?

from https://wordpress.org/support/topic/database-bloat-unable-to-archive-records/
> We’re trying to perform database optimisation, we’re seen the Stream database is taking up a lot of space :
wp_stream_meta 1347.50 (MB)
> 
> We’ve tried to optimise this by adjusting the days of records kept from 30 days to 5 days and the database size remains the same.
> 
> We’ve confirmed the wp_stream_auto_purge is running and not getting killed.
> 
> We’ve also attempted to reset the Stream database via Advanced tools in the plugin and it does not work, the page reloads but all the records prior to the action remain in the DB.
> 
> There seems to be a problem with the plugin – please advise.
> 
> Site Info:
> WP: V6.2.2
> php: 8.0.29
> mysql: 5.7.42-46-log

Issue #1476, opened 2024-04-02
Export stops at 10,000 items


cbirdsong:
# Bug Report

## Expected Behavior

A CSV export should contain all the data Stream has.

## Actual Behavior

The export is cut off at 10,000 items.

## Steps to Reproduce the Problem

1. Set a high-traffic site to retain data for enough time to log more than 10,000 items.
2. Wait until there are more than 10,000 items.
3. Export to CSV or JSON
4. Open the exported file. You should find only 10,000 rows or items.

I have only seen this on one site

## System Information

- Stream plugin version: 4.0.0
- WordPress version: 6.4.3
- PHP version: 7.4
- Browser: Firefox
- Computer operating system: Mac OS

---

This would be largely sidestepped by fixing #1191

cbirdsong:
It looks like this should be an easy fix:

```php
/**
* Increase pagination limit for CSV Output
*
* @param int $records_per_page Old limit for records_per_page.
* @return int
*/
public function disable_paginate( $records_per_page ) {
    return 10000;
}
```

I have no idea what a sensible number to cap it at is, though. The site I discovered this on has 164,000 records and counting and Stream otherwise works _fine_, so probably higher than that?

lkraav:
> I have no idea what a sensible number to cap it at would be, though. The site I discovered this on has 164,000 records and counting and Stream otherwise works _fine_ with that volume, so probably higher than that?

`apply_filters()` makes sense here, so everybody can have it according to their environment capability.

PR #1531, opened 2024-07-30
Allow commas in CSV exporter fields

### LINKED ISSUE #1416: CSV export is malformed when any value contains a comma
# Bug Report

## Expected Behavior

When exporting logs as a CSV, all data should be well-formed.

## Actual Behavior

When any field value contains a comma, the CSV is malformed.

1. Set your user display name to one that contains a comma, for example "Last Name, First Name"
2. Perform an action that gets logged in the audit log
3. Visit the audit log screen and download a CSV from the Record Actions menu at the bottom
4. Open the CSV in a spreadsheet application which will show the data in columns
5. Observe that the CSV output for the affected row is malformed due to the unquoted comma

## Suggested Fix

The `Exporter_CSV::output_file()` method should use `fputcsv()` instead of manually concatenating fields with commas.

## System Information

- Stream plugin version: 3.9.2
- WordPress version: 6.1
- PHP version: 8.0

tharsheblows:
Fixes #1416 

Use `fputcsv()` to format csv and write to file. 

# Checklist

- [x] I have ~added~ updated phpunit tests.

## Release Changelog

- Fix: commas no longer break the csv exporter


exporters/class-exporter-csv.php
```diff
@@ -38,12 +38,18 @@ public function output_file( $data, $columns ) {
...
+		$csv = fopen( 'php://output', 'w' );
+		fputcsv( $csv, array_values( $columns ) );
+
 		foreach ( $data as $row ) {
-			$output .= join( ',', $row ) . "\n";
+			fputcsv( $csv, $row );
 		}
+		fclose( $csv ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_operations_fclose
+
+		echo ob_get_clean(); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
tharsheblows, line 51: @delawski I'm worried about the size of the output buffer, do you think I need to be?

tharsheblows: See https://github.com/xwp/stream/issues/1476 and https://github.com/xwp/stream/issues/1503 for some large log sizes

delawski: That's a good question. I guess that with a high enough number of records we may hit a PHP timeout limit (depending on the server configuration). However, some pro/enterprise users may still want to export everything.

I think a reasonable short-term solution to that issue would be to allow setting the number of records to export (as suggested [here](https://github.com/xwp/stream/issues/1476#issuecomment-2034374922)). We could default to 10000 records but make that value filterable.

A more sophisticated solution (maybe for the future) would be to rely on some async queue (Action Scheduler?) that would export the records to a file stored in the filesystem. Once everything is exported, a link to that file would be available in WP admin.

What are your thoughts?

delawski (approved):

exporters/class-exporter-csv.php
tharsheblows: > I think a reasonable short-term solution to that issue would be to allow setting the number of records to export (as suggested https://github.com/xwp/stream/issues/1476#issuecomment-2034374922). We could default to 10000 records but make that value filterable.

I'm not sure how to handle pagination though? I guess a first step is allowing that to be filterable and then figuring out if / where people are hitting the php limits (and testing that ourselves!)

> A more sophisticated solution (maybe for the future) would be to rely on some async queue (Action Scheduler?) that would export the records to a file stored in the filesystem. Once everything is exported, a link to that file would be available in WP admin.

I like this, it should be straightforward although we'd need to think about security but we could also make it be deleted after x amount of time too. (I 💖 Action Scheduler)

tharsheblows: @delawski Yes, that makes sense. There's an [open issue](https://github.com/xwp/stream/issues/1476) for filtering the number. The WP CLI command has a `records_per_page` field that can be used to get an arbitrary number of records.

PR #1533, opened 2024-07-30
add port so we can see the database

tharsheblows:
Add in ports in `docker-compose.yml` to allow viewing of the database in, eg, Sequel Ace 😊

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,6 +13,17 @@ services:
       MYSQL_PASSWORD: password
       MYSQL_ROOT_PASSWORD: password
 
+  phpmyadmin:
+    image: phpmyadmin/phpmyadmin:5.2.1
+    ports:
+      - "8080:80"
+    depends_on:
+      - mysql
+    environment:
+      PMA_HOST: mysql
+      PMA_USER: root
+      PMA_PASSWORD: password
+
   wordpress:
     image: ghcr.io/xwp/stream-wordpress:${WORDPRESS_IMAGE_VERSION:-latest}
     build:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,7 +14,7 @@ services:
       MYSQL_ROOT_PASSWORD: password
 
   phpmyadmin:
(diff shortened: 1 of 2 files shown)


delawski (approved):
Looks good, thank you!

I think we could also add a phpmyadmin container to have a more integrated and dependency-free way of interacting with the DB, something like:

```yaml
  phpmyadmin:
    image: phpmyadmin/phpmyadmin:5.2.1
    depends_on:
      - db
    environment:
      VIRTUAL_HOST: phpmyadmin.${DEV_URL:-stream.wpenv.net}
      PMA_HOST: wordpress
      PMA_USER: wordpress
      PMA_PASSWORD: password
```

What's your take?

tharsheblows:
@delawski 

> I think we could also add a phpmyadmin container ...

Yes! It's available at http://stream.wpenv.net:8080 now. Due to multisite shenanigans, it didn't seem possible as a subdomain😊

Could you test that it works for you?

docker-compose.yml
```diff
@@ -4,13 +4,26 @@ services:
...
+      - "3306:3306"
     restart: always
     environment:
       MYSQL_DATABASE: wordpress
       MYSQL_USER: wordpress
       MYSQL_PASSWORD: password
       MYSQL_ROOT_PASSWORD: password
 
+  phpmyadmin:
+    image: phpmyadmin/phpmyadmin:5.2.1
```
delawski, line 17: **🧹 chore:** So I tested it locally and while phpmyadmin worked well and was available on the port 8080, I noticed a platform mismatch warning in console when pulling the Docker image.

It was strange because phpmyadmin on Docker Hub is available for many platforms, including mine, `linux/arm64/v8`:


Screenshot 2024-07-31 at 17 41 59

It turned out that when I changed the `image` value as follows, everything started to works seamlessly:

```suggestion
    image: phpmyadmin:5.2.1
```

I'm sorry that I provided an incorrect suggestion in the first place 😅

tharsheblows: Oooh good catch! I even checked dockerhub and it didn't register! I've committed it and am going to merge.

PR #1543, opened 2024-08-05
Use scheduled action for resetting large record and meta tables

### LINKED ISSUE #1503: Large tables not able to be reset

tharsheblows:
Fixes #1503 

# Development notes

## Changes to default database reset functionality:

 - if the site is a single site installation or multisite install with the plugin network activated, the Stream and Stream meta tables are truncated and then a cleanup function is run on the meta table. (This is for the case where a record is added in between truncation the `{$wpdb->streammeta}` table and the `{$wpdb->stream}` table.)
 - if the site is a multisite which is not network activated and
   - the site has <= 1000000 records in the `{$wpdb->stream}` table: the previous deletion functionality will run
   - the site has > 1000000 records in the `{$wpdb->stream}` table:
     - an action will be scheduled async to delete 250k records at a time using the previous deletion functionality. The limit is handled by a range on the `{$wpdb->stream}.ID` column
     - the action is recursive and will run until there are no records left with an ID lower than the initial ID. This is to avoid the situation where records keep getting added as the actions run so they never finish.

The "Reset Stream Database" link in Stream --> Settings --> Advanced will not show while the records are being deleted. The warning given will depend on whether or not it's going to be an asynchronous process.

In my local environment, each pass of deleting 250k records takes 10-15 seconds. 

## Added filters

- `wp_stream_is_large_records_table`: filter the number of records to be processed when deciding what is a large records table. Default is 1M, ie anything over 1M records to be processed is considered a large records table.
- `wp_stream_batch_size`: how many records in the `{$wpdb->stream}` table will be used when doing the deletion.

## Incidental additions:
 - `$this->plugin->get_site_type()` returns what sort of site it is:
  - `Admin::WP_STREAM_MULTI_NETWORK` is a multisite with the plugin network activated
  - `Admin::::WP_STREAM_MULTI_NOT_NETWORK` is a multisite where the plugin is not network enabled
  - `Admin::WP_STREAM_SINGLE_SITE` is a single site
- [Action Scheduler library ](https://github.com/woocommerce/action-scheduler) to use instead of WP cron functions

## Testing instructions:

You can use the `npm run large-records-generate` to generate a bunch of records. You can see how many you've generated by running `npm run large-records-show`

# Checklist

## Release Changelog

- Fix: Use async process for resetting the Stream database on large multisites where the plugin is not network enabled
- Fix: Truncate tables when resetting the Stream database on single sites and multisites where the plugin is network enabled


classes/class-admin.php
```diff
@@ -728,19 +759,165 @@ public function wp_ajax_reset() {
...
+				$last_entry + 1, // A tweak to get it correct the first time through.
+				get_current_blog_id()
+			)
+		);
+
+		if ( empty( $start_from ) ) {
+			return;
+		}
+
+		// We will do at most 500000 at a time.
```
delawski, line 867:
```suggestion
		// We will do at most 250000 at a time.
```

classes/class-admin.php
```diff
@@ -728,19 +759,165 @@ public function wp_ajax_reset() {
 	private function erase_stream_records() {
 		global $wpdb;
 
-		$where = '';
+		// If this is a multisite and it's not networked activated,
```
delawski, line 762:
```suggestion
		// If this is a multisite and it's not network activated,
```

classes/class-admin.php
```diff
@@ -18,6 +18,26 @@
...
+
+	/**
+	 * Used to check if it's a multisite with the plugin network enabled.
+	 */
+	const WP_STREAM_MULTI_NETWORK = 'multisite-network';
+
+	/**
+	 * Used to check if it's a multisite with the plugin not network enabled.
+	 */
+	const WP_STREAM_MULTI_NOT_NETWORK = 'multisite-not-network';
```
delawski, line 34: **💡 thought:** Two things, nitpicks really:
1. I feel like this belongs to the `Plugin` class and not `Admin` (alternatively, this could be part of some kind of a `Utility` class). What's your take?
2. I guess it's a matter of preference but I wonder what do you think about having a set of methods for checking the installation type instead of those enum-like string constants? Something like `Admin::is_single_site()`, `Admin::is_multisite_network_activated()` and `Admin::is_multisite_not_network_activated()`.
It would follow the WordPress convention like `is_network_admin()` and make the conditions easier to read (potentially). But then they couldn't be used in a `switch` statement later on and a regular `if ... else` would have to be used.

classes/class-settings.php
```diff
@@ -471,6 +471,50 @@ public function get_defaults() {
...
+	private function get_async_deletion_warning() {
+		// Check if there is an action scheduler event running already deleting things.
+		$has_scheduled_action = Admin::is_running_async_deletion();
+
+		return $has_scheduled_action ?
+			__( 'Currently deleting records. Please be patient, this can take a while.', 'stream' )
+			:
+			__( 'Warning: This will delete all activity records from the database for this site. If any are added while this is running, they will not be deleted.', 'stream' );
+	}
+
```
delawski, line 517: **🥜 nitpick:** A very minor thing yet I'd like to share with you. I find this code snippet relatively hard to read because of this ternaries and indirection between them (one ternary calls a method which has yet another ternary inside). Do you think it'd be clearer if the `get_async_deletion_warning()` logic was moved over to the switch statement and the regular `if .. else` statements used instead of ternaries?

delawski (approved):
The code looks good to me 👍

classes/class-admin.php
bartoszgadomski: I agree with @delawski, but also have one comment about the constants naming:

I'm not sure what's the naming convention in this plugin, but addition of the `WP_STREAM_` prefix seems to not be necessary here (these constants are scoped in `Admin` class).

classes/class-admin.php
```diff
@@ -728,19 +759,165 @@ public function wp_ajax_reset() {
 	private function erase_stream_records() {
 		global $wpdb;
 
-		$where = '';
+		// If this is a multisite and it's not networked activated,
+		// only delete the entries from the blog which made the request.
+		if ( self::WP_STREAM_MULTI_NOT_NETWORK === $this->plugin->get_site_type() ) {
```
delawski, line 764: **❓question:** I guess my only concern here is that the user may still want to erase Stream data of a single site even if Stream is network-activated.

The reason is that in order to trigger the data removal, you need to go to a specific site -> Stream -> Settings and click that button there. It may be misleading (even if we have the warning message) that triggering the removal on that particular site will affect the entire network. Also, a "regular" admin on a child site may be able to initiate a destructive action for the entire network (which only super-admins should be able to do).

I think we may need to make that action more specific and by default only remove single site data, even if Stream is network activated. On top of that, we would either need:
- a network-wide Stream Settings page where the entire network data removal would be possible, or
- a second button on the current Settings page which would delete data from the entire network and would only be displayed for super-admins.

Are my concerns valid?

classes/class-admin.php
```diff
@@ -728,19 +759,165 @@ public function wp_ajax_reset() {
...
 
-		if ( is_multisite() && ! $this->plugin->is_network_activated() ) {
-			$where .= $wpdb->prepare( ' AND `blog_id` = %d', get_current_blog_id() );
+			// First check the log size.
+			$stream_log_size = self::get_blog_record_table_size();
+
+			// If this is a large log and we need to delete only the entries
+			// pertaining to an individual site, we will need to do those in batches.
+			if ( $this->plugin->is_large_records_table( (int) $stream_log_size ) ) {
```
bartoszgadomski, line 771: @tharsheblows The `get_blog_record_table_size` method always returns `int`, there's no need for a cast here:

```suggestion
			if ( $this->plugin->is_large_records_table( $stream_log_size ) ) {
```

classes/class-admin.php
```diff
@@ -728,19 +759,165 @@ public function wp_ajax_reset() {
...
+				return;
+			}
+
+			$wpdb->query(
+				$wpdb->prepare(
+					"DELETE `stream`, `meta`
+					FROM {$wpdb->stream} AS `stream`
+					LEFT JOIN {$wpdb->streammeta} AS `meta`
+					ON `meta`.`record_id` = `stream`.`ID`
+					WHERE 1=1 AND `blog_id`=%d;",
```
bartoszgadomski, line 782: @tharsheblows The `WHERE 1=1` is typically added when a SQL query is dynamically generated. While it will be ignored by the database, there's no need to have it in code in this case, as all the "where" clauses are static:

```suggestion
					`blog_id`=%d;",
```

classes/class-settings.php
```diff
@@ -471,6 +471,50 @@ public function get_defaults() {
...
+		switch ( $site_type ) {
+			case Admin::WP_STREAM_MULTI_NETWORK:
+				$warning = __( 'Warning: This will delete all activity records from the database for all sites.', 'stream' );
+				break;
+			case Admin::WP_STREAM_MULTI_NOT_NETWORK:
+				$warning = $this->plugin->is_large_records_table( Admin::get_blog_record_table_size() ) ?
+						$this->get_async_deletion_warning() :
+						__( 'Warning: This will delete all activity records from the database for this site.', 'stream' );
+				break;
```
delawski, line 491: **❗️issue:** I think I found a bug here. I've got an async deletion scheduled and underway. However, on the Settings page I'm getting the latter warning message:

Screenshot 2024-08-06 at 14 36 17

I think it's because we're checking if it's a large records table based on _current_ table site, i.e.:

```php
$this->plugin->is_large_records_table( Admin::get_blog_record_table_size() )
```

As soon as the async tasks remove bunch of records and the `get_blog_record_table_size()` goes below the threshold set in `is_large_records_table()` method, we won't be seeing the `get_async_deletion_warning()` but rather the other warning message (even though an async tasks are being run in the "background").

classes/class-plugin.php
```diff
@@ -113,6 +113,9 @@ public function __construct() {
 
 		spl_autoload_register( array( $this, 'autoload' ) );
 
+		// Load Action Scheduler.
+		require_once $this->locations['dir'] . '/vendor/woocommerce/action-scheduler/action-scheduler.php';
```
bartoszgadomski, line 117: @tharsheblows There's a risk of conflict if this package was loaded by another plugin. Should this be preceded with `function_exists` call (or something similar) to only load this class if it has not been loaded by another plugin earlier?

classes/class-plugin.php
```diff
@@ -333,4 +336,32 @@ public function is_mustuse() {
...
+	}
+
+	/**
+	 * Should the number of records which need to be processed be considered "large"?
+	 *
+	 * @param int $record_number The number of rows in the {$wpdb->prefix}_stream table to be processed.
+	 * @return bool Whether or not this should be considered large.
+	 */
+	public function is_large_records_table( int $record_number ): bool {
+		return apply_filters( 'wp_stream_is_large_records_table', $record_number > 1000000, $record_number );
```
bartoszgadomski, line 365: @tharsheblows Suggestion: adding a docblock comment to all custom hooks would be beneficial for developers to understand how this plugin can be adjusted :)

classes/class-admin.php
tharsheblows: > I feel like this belongs to the Plugin class and not Admin (alternatively, this could be part of some kind of a Utility class). What's your take?
Yes! I agree. I'll put it in the `Plugin` class for now and if it feels out of place there, move to a `Utility` class.

> I guess it's a matter of preference but I wonder what do you think about having a set of methods for checking the installation type instead of those enum-like string constants? Something like Admin::is_single_site(), Admin::is_multisite_network_activated() and Admin::is_multisite_not_network_activated().

Yes, I like this also, it makes sense to me and would make development easier.

tharsheblows: @bartoszgadomski Yes! You're right, that prefix isn't necessary and also not in line with the naming condition! I'll fix those.

tharsheblows: Commit : https://github.com/xwp/stream/pull/1543/commits/a6d2d36f9b95aaf0a0357e90051b25066b1ce4d3

If I missed doing something in that, it's accidental so please tell me

classes/class-admin.php
tharsheblows: Commit https://github.com/xwp/stream/pull/1543/commits/e931533237950d63a23b67adfcd7870fcd955ed6 

Thank you! :)

classes/class-plugin.php
tharsheblows: @bartoszgadomski Action Scheduler works when included multiple times! It's a clever library – it loads the most recent version requested so all of the functions expected are available. Eg latest version does it [like this ](https://github.com/woocommerce/action-scheduler/blob/trunk/action-scheduler.php)

classes/class-plugin.php
tharsheblows: @bartoszgadomski ACK yes, I totally forgot.

tharsheblows: Commit https://github.com/xwp/stream/pull/1543/commits/611129575f69028f0d6074640abf42bf676331e1

composer.json
```diff
@@ -12,8 +12,9 @@
   ],
   "require": {
     "php": ">=7.0",
+    "ext-json": "*",
     "composer/installers": "~1.0",
-    "ext-json": "*"
+    "woocommerce/action-scheduler": "^3.8"
```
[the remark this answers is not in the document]
tharsheblows: @delawski I think it'd be the same in the scheduled purges? It needs to run as a background process, I'm not sure we can account for anyone who's turned off cron or has such a small site that it doesn't run? I think maybe in that case there wouldn't be a lot of logs so it wouldn't run as a background process?

I'm not sure how to account for that case. We're using a recursive background process because it wouldn't necessarily work any other way so I don't think they'll be worse off. ?? That logic may be faulty.

classes/class-settings.php
tharsheblows: @delawski I've updated it here to use `if ... else` statements and checked for the AS action first as I think that's the most robust for future development (in case anything else uses the async deletion, we don't have to worry about the conditional here) plus it's a cheap check. Is that ok?

Also I added more empty lines than usual as I have a bit of trouble seeing if / else statements when there are no empty lines for some reason.

Commit https://github.com/xwp/stream/pull/1543/commits/12ca1909854decc9e0326998e4cf884f890ea487

classes/class-admin.php
tharsheblows: @delawski 

> I guess my only concern here is that the user may still want to erase Stream data of a single site even if Stream is network-activated.

This would be adding a feature which is not currently there. We could do that but we would need some more checks to make sure we're not trying to delete the same thing more than once? I'd kind of prefer to put this in either the next release or in a separate issue for ease of backing out if it needs to be.

> The reason is that in order to trigger the data removal, you need to go to a specific site -> Stream -> Settings and click that button there.

When the plugin is network activated, the only place to remove data is in the network admin. The button to remove data isn't in the individual sites. And vice versa -- when the plugin is not network activated, it doesn't show up in the network admin. 

On your local environment, the `is_network_activated()` check doesn't work correctly due to the rewriting of `stream` to `stream-src`, I need to make an issue for that but keep forgetting. I have a mu plugin with that filter which I use to test it being network activated:
`add_filter( 'wp_stream_is_network_activated', '__return_true' );`

I should probably write an e2e test for the two multisite possibilities? Although that would require fixing that rewrite issue.

classes/class-settings.php
tharsheblows: @delawski YES you're right -- I've changed the logic in https://github.com/xwp/stream/pull/1543/commits/12ca1909854decc9e0326998e4cf884f890ea487 to check if there's an async deletion running first which should avoid this issue.

classes/class-settings.php
delawski: This looks great, thank you!

> Also I added more empty lines than usual as I have a bit of trouble seeing if / else statements when there are no empty lines for some reason.

All good! 😄 I wonder if it would make it easier to read for you if there were only `if` statements (without any `elseif`) and they simply returned the warning message directly (there's no other logic besides that `if .. else` in the function). Anyway, let's leave it as is, it's super clean and readable to me now 👍

classes/class-admin.php
delawski: > I'd kind of prefer to put this in either the next release or in a separate issue for ease of backing out if it needs to be.
💯 agree on that - let's not work on it in this release.

Also, in light of what you just explained to me:

> When the plugin is network activated, the only place to remove data is in the network admin. The button to remove data isn't in the individual sites. And vice versa -- when the plugin is not network activated, it doesn't show up in the network admin.

it may not be necessary to add such a feature at all. The user flows are correct and messaging is clear 👍

composer.json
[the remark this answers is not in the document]
delawski: > That logic may be faulty.

No, I think your logic is correct. Only some very small (DIY?) hosting services would not have CRON running. I wouldn't assume such servers host sites that would require Stream at all.

Something else I could have considered before posting that question is that Woo itself relies on Action Scheduler and CRON. If they are okay with that choice, we should be too.

delawski (approved):
Thank you for addressing the feedback. The PR looks fabulous!

bartoszgadomski (approved):

PR #1547, opened 2024-08-07
Fix local environment path

### LINKED ISSUE #1545: Fix local environment check on the plugin being network activated
# Bug Report

Due to the rewrite of `stream` to `stream-src` in the local environment, the `is_network_activated()` check fails.

## Expected Behavior

When the plugin is network activated, `is_network_activated()` should return true.

## Actual Behavior

It returns false.

1. Activate the plugin for the network
2. Observe that Stream does not show up on the network admin menu and all the checks for `is_network_activated()` fail.

tharsheblows:
Fixes #1545 

Fixes the path to the plugin file and adds e2e tests which show the network admin UI in the network activated and deactivated states.

It also introduces a setup step for the e2e tests which logs a user in as admin and deactivates the plugin so we know which state we're in when starting the tests.

To run all of the e2e tests, use `npm run tests-e2e`

# Checklist

## Release Changelog

- Fix: `$this->plugin->is_network_activated()` works correctly now on local environments.

### DIFF
--- a/tests/e2e/network-activated.spec.js
+++ b/tests/e2e/network-activated.spec.js
@@ -1,0 +1,56 @@
+/**
+ * WordPress dependencies
+ */
+import { test, expect } from '@wordpress/e2e-test-utils-playwright';
+
+let page;
+
+test.describe.configure( { mode: 'serial' } );
+
+test.beforeAll( async ( { browser } ) => {
+	page = await browser.newPage();
+} );
+
+test.afterAll( async () => {
+	await page.goto( 'http://stream.wpenv.net/wp-admin/network/plugins.php' );
+	const isActive = await page
+		.getByLabel( 'Network Deactivate Stream' )
+		.isVisible();
+
+	if ( isActive ) {
+		// eslint-disable-next-line no-console
+		console.log( 'Deactivating Stream after Network tests.' );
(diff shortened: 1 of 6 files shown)


tests/e2e/network-activated.spec.js
```diff
@@ -0,0 +1,56 @@
...
+		// Expects Stream log to not have the Network Settings.
+		await expect(
+			page.locator( '[href*="admin.php?page=wp_stream"]' ),
+		).not.toBeVisible();
+	} );
+
+	// We should not have an updated row. This times out which makes it fail.
+	test( 'does not have updated row', async () => {
+		await page.goto( 'http://stream.wpenv.net/wp-admin/network/plugins.php' );
+		const isActive = await page
```
delawski, line 40: **🧹 chore:** I believe this should be the opposite:

```suggestion
		const isInactive = await page
```

tests/e2e/network-activated.spec.js
```diff
@@ -0,0 +1,56 @@
...
+			.getByLabel( 'Network Activate Stream' )
+			.isVisible();
+
+		if ( isActive ) {
```
delawski, line 44:
```suggestion
		if ( isInactive ) {
```

delawski (approved):
Looks great! I left just a minor suggestion and I'm approving already.

PR #1549, opened 2024-08-08
Fix/1222 log post author change

### LINKED ISSUE #1222: Change of post author is not recorded
Delete the section that is not applicable:

# Bug Report

## Expected Behavior

When a post is edited and only the post author is changed, this change should be recorded by Stream.

## Actual Behavior

No log is recorded at all.

1. Edit a post
2. Change the author
3. Save the post
4. Check the Stream log and observe there's no record of the change

## System Information

- Stream plugin version: 3.5.1
- WordPress version: 5.6
- PHP version: 7.4

tharsheblows:
Fixes #1222 

This changes the hook used for watching for post updates to `post_updated` and uses `set_object_terms` to watch for term changes. The idea is that everything which triggered or should have triggered a log previously still triggers a log but with more information.

The strings in the summary are built separately and then the data itself is stored in an associative array in the meta table. (nb: this should be a JSON object for cross language portability I think)

Currently tests are failing as I'm not convinced this is completely the right approach.

# Checklist


connectors/class-connector-posts.php
```diff
@@ -24,8 +26,9 @@ class Connector_Posts extends Connector {
 	 * @var array
 	 */
 	public $actions = array(
-		'transition_post_status',
 		'deleted_post',
+		'post_updated',
```
delawski, line 30: **💡 thought:** I believe this will not log new posts creation because `post_updated` is [only called for existing posts](https://github.com/WordPress/wordpress-develop/blob/6.6/src/wp-includes/post.php#L4860-L4860).

It may be a good thing because I think we don't need to log e.g. a `post_title` change for a new post. However, without `transition_post_status` callback we no longer log new posts creation. I guess we could use e.g. [`save_post`](https://github.com/WordPress/wordpress-develop/blob/6.6/src/wp-includes/post.php#L4891) action or bring back `transition_post_status` but limit its scope to only log transitions to the `publish` status.

tharsheblows: @delawski `$update` is [a check on whether or not the data has a `post_id`](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/post.php#L4254) which gets set immediately in the editor but ... if someone is using `wp_insert_post()` directly to add a post then it won't so I need to add that in, thank you!

tharsheblows: @delawski I've changed this to use `wp_after_insert_post` -- it's noisier but because of the orders of the filters in the REST API, it's the one that works there as well as in the Quick Edit section.

I still need to test and add automated tests though!

PatelUtkarsh:
Hi @tharsheblows, are you still planning to finish this? #1222 is still a real bug and we'd like to see it land.

If you want to continue, three things need to happen before merge:

1. Rebase onto current `develop`
2. PHPUnit coverage for author change, term add/remove, post_content edit, and the "no duplicate logs" case.

Happy to take it over if you'd rather hand it off. Let us know.

PR #1587, opened 2024-09-20
Reintroduce deployment to packagist.org

### LINKED ISSUE #1569: add back automation to push built plugin to dist repo
# Bug Report

In 4.0.0, the Travis CI workflow which ported the built plugin to [the distribution repo](https://github.com/xwp/stream-dist) broke. It was removed in 4.0.1. The distribution repo is tied to [the Stream package in Packagist ](https://packagist.org/packages/xwp/stream) and needs to be kept up to date with releases.

nb: this could be a tooling only update and if it is released on its own, may not need pushing to the WordPress plugins repo, although this should be discussed.

## Expected Behavior

Releases should appear on Packagist via the distribution repo.

## Actual Behavior

Releases are not pushed to the distribution repo.

1. Look at the 4.0.1 release in this repo
2. Notice that it does not appear in the distribution repo or on Packagist.

marcinkrzeminski:
Fixes #1569

This PR reintroduces deployment to the read-only `xwp/stream-dist` repository. That repository is connected with our packagist.org account to make our plugin available via the Packagist repository.


.github/workflows/deploy-to-stream-dist.yml
```diff
@@ -0,0 +1,129 @@
+name: Deploy to xwp/stream-dist Repository
+
+on:
+  release:
+    types:
+      - released
```
tharsheblows, line 6: **❓question**
Will this run on releases only now? What do you think about keeping `master` and maybe `develop` up to date on the dist repo and Packagist? I know it's under discussion.

tharsheblows (commented):
@marcinkrzeminski I've left one comment which we're discussing on Slack, it's about what should go to the dist repo.

.github/workflows/deploy-to-stream-dist.yml
marcinkrzeminski: @tharsheblows, my thought process here was to consider a scenario where Packagist equals wp.org. Since the release branches are based on `develop`, which in turn is based on `master`, we should be covered. I reviewed the Release Cycle to reach this conclusion: https://github.com/xwp/stream/blob/develop/contributing.md#release-cycle. Let me know your thoughts.

tharsheblows (approved):
I left a comment but it's not a blocker!

PR #1616, opened 2024-10-22
fix: ESLint issues preventing GitHub Actions to run

marcinkrzeminski:
This PR removes obsolete / redundant ESLint dependency. Since we're using @wordpress/scripts that's installing ESLint internally we don't need to maintain ourselves. What's more important this fixes an issue with broken lint GH action which was reporting issues related to the latest version of ESLint which was updated constantly by the dependbot.


pereirinha (approved):
@marcinkrzeminski 

Thanks for this. I just tested localy and it's working as expected.
The only thing missing is resolving the merge conflict on the lock file.

PR #1656, opened 2024-12-19
Use WC_Order_Factory  to get order id

### LINKED ISSUE #1655: Fatal error on attempt to empty the trash for the WooCommerce orders
# Bug Report

https://wordpress.org/support/topic/conflict-with-the-woocommerce-plugin/

## Expected Behavior

Describe what should happen.

## Actual Behavior

Attempted to empty the trash for the WooCommerce orders and it results in a fatal error. Below is the error that it logs.

```
#3 /wp-content/plugins/stream/connectors/class-connector-woocommerce.php(440): WC_Abstract_Order->__construct()

#4 [internal function]: WP_Stream\Connector_Woocommerce->callback_deleted_post()

#5 /wp-content/plugins/stream/classes/class-connector.php(127): call_user_func_array()

#6 /wp-includes/class-wp-hook.php(324): WP_Stream\Connector->callback()
```

1.
2.
3.

## Screenshots

Drag and drop screenshots in this area that help illustrate the problem.

## System Information

WordPress and all plugins are up to date and the site and I’ve tested it with the TwentyTwenyFour theme. The empty trash function works as expected if the Stream plugin is deactivated.

- Stream plugin version:
- WordPress version:
- PHP version:
- Browser:
- Computer operating system:

marcinkrzeminski:
Fixes #1655

# Checklist

## Release Changelog

- Fix: Use preferred WC_Order_Factory to get the order ID

### DIFF
--- a/connectors/class-connector-woocommerce.php
+++ b/connectors/class-connector-woocommerce.php
@@ -437,8 +437,8 @@ class Connector_Woocommerce extends Connector {
 			return;
 		}
 
-		$order           = new \WC_Order( $post->ID );
-		$order_title     = esc_html__( 'Order number', 'stream' ) . ' ' . esc_html( $order->get_order_number() );
+		$order_id        = \WC_Order_Factory::get_order_id( $post->ID );
+		$order_title     = esc_html__( 'Order number', 'stream' ) . ' ' . esc_html( $order_id );
 		$order_type_name = esc_html__( 'order', 'stream' );
 
 		$this->log(
--- a/connectors/class-connector-woocommerce.php
+++ b/connectors/class-connector-woocommerce.php
@@ -396,8 +396,8 @@ class Connector_Woocommerce extends Connector {
 			$action = 'updated';
 		}
 
-		$order           = new \WC_Order( $post->ID );
-		$order_title     = esc_html__( 'Order number', 'stream' ) . ' ' . esc_html( $order->get_order_number() );
+		$order_id        = \WC_Order_Factory::get_order_id( $post->ID );
+		$order_title     = esc_html__( 'Order number', 'stream' ) . ' ' . esc_html( $order_id );
 		$order_type_name = esc_html__( 'order', 'stream' );
 
 		$this->log(
@@ -495,8 +495,8 @@ class Connector_Woocommerce extends Connector {
 			'stream'
 		);
 
-		$order           = new \WC_Order( $order_id );
-		$order_title     = esc_html__( 'Order number', 'stream' ) . ' ' . esc_html( $order->get_order_number() );
+		$order_id        = \WC_Order_Factory::get_order( $order_id );
+		$order_title     = esc_html__( 'Order number', 'stream' ) . ' ' . esc_html( $order_id );
 		$order_type_name = esc_html__( 'order', 'stream' );
 
 		$this->log(
--- a/classes/class-connector.php
+++ b/classes/class-connector.php
@@ -146,12 +146,12 @@ abstract class Connector {
 	/**
 	 * Log handler
 	 *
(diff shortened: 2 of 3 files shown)


connectors/class-connector-woocommerce.php
```diff
@@ -437,8 +437,8 @@ public function callback_deleted_post( $post_id ) {
 			return;
 		}
 
-		$order           = new \WC_Order( $post->ID );
-		$order_title     = esc_html__( 'Order number', 'stream' ) . ' ' . esc_html( $order->get_order_number() );
+		$order_id        = \WC_Order_Factory::get_order_id( $post->ID );
```
tharsheblows, line 440: **❓question**
I'm not sure on this -- `WC_Order_Factory::get_order_id()` will always return the post ID: 

```	
        /**
	 * Get the order ID depending on what was passed.
	 *
	 * @since 3.0.0
	 * @param  mixed $order Order data to convert to an ID.
	 * @return int|bool false on failure
	 */
	public static function get_order_id( $order ) {
		if ( false === $order ) {
			return self::get_global_order_id();
		} elseif ( is_numeric( $order ) ) {
			return $order;
		} elseif ( $order instanceof WC_Abstract_Order ) {
			return $order->get_id();
		} elseif ( ! empty( $order->ID ) ) {
			return $order->ID;
		} else {
			return false;
		}
	}
```

Is that the order ID now? 

Although -- this is on the `deleted_post` hook so it should be ok?

tharsheblows (changes requested):
There's another occurrence of this in [the post status transition hook here](https://github.com/xwp/stream/blob/develop/connectors/class-connector-woocommerce.php#L399-L401) -- should that one be updated also?

connectors/class-connector-woocommerce.php
marcinkrzeminski: @tharsheblows good question, however I agree that we should be OK since this action is triggered on post deletion. 

While I was working on a solution, I came across this comment on the [abstract-wc-order.php](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/abstracts/abstract-wc-order.php#L111-L118), and that's why I used it.

```
* Get the order if ID is passed, otherwise the order is new and empty.
* This class should NOT be instantiated, but the wc_get_order function or new WC_Order_Factory
* should be used. It is possible, but the aforementioned are preferred and are the only
* methods that will be maintained going forward.
 ```

**Comment by marcinkrzeminski** [comment]
> There's another occurrence of this in [the post status transition hook here](https://github.com/xwp/stream/blob/develop/connectors/class-connector-woocommerce.php#L399-L401) -- should that one be updated also?

Good catch. I've found another place and updated it as well to use the `\WC_Order_Factory::get_order_id()`

File: connectors/class-connector-woocommerce.php (hunk 1 above)
Comment (tharsheblows) [reply connectors/class-connector-woocommerce.php:440]: @marcinkrzeminski Ah ha, thanks!

File: connectors/class-connector-woocommerce.php [hunk 2]
```diff
@@ -495,8 +495,8 @@ public function callback_woocommerce_order_status_changed( $order_id, $old_order
 			'stream'
 		);
 
-		$order           = new \WC_Order( $order_id );
-		$order_title     = esc_html__( 'Order number', 'stream' ) . ' ' . esc_html( $order->get_order_number() );
+		$order_id        = \WC_Order_Factory::get_order( $order_id );
+		$order_title     = esc_html__( 'Order number', 'stream' ) . ' ' . esc_html( $order_id );
```
tharsheblows, line 499: **👏 praise**
Nice catch!

tharsheblows (approved):

PR #1658, opened 2024-12-20
Theme Edits Grayed Out

### LINKED ISSUE #1657: Theme Edits Grayed Out
# Bug Report

https://wordpress.org/support/topic/theme-edits-grayed-out/

## Expected Behavior

Describe what should happen.

## Actual Behavior

There were suspicious changes made to the functions.php file and when I went to Stream to see who had done it there was nothing there……not even the fix I had just made. So when I went to filter the streams by Theme Edits I notice it is grayed out – there is no option to view those. There is also nowhere in the Settings I can enable it.

Why is this and how can I see that. You would think that would be a major edit admins would want to be able to see

1.
2.
3.

## Screenshots

eRElbJ

## System Information

- Stream plugin version:
- WordPress version:
- PHP version:
- Browser:
- Computer operating system:

marcinkrzeminski:
Fixes #1657

Use alternative hook to track the theme files updates in the Dashboard.

# Checklist

## Release Changelog

- Fix Theme files changes are tracked as expected

### DIFF
--- a/connectors/class-connector-editor.php
+++ b/connectors/class-connector-editor.php
@@ -46,8 +46,8 @@ class Connector_Editor extends Connector {
 	 */
 	public function register() {
 		parent::register();
-		add_action( 'load-theme-editor.php', array( $this, 'get_edition_data' ) );
-		add_action( 'load-plugin-editor.php', array( $this, 'get_edition_data' ) );
+
+		add_action( 'wp_ajax_edit-theme-plugin-file', array( $this, 'get_edition_data' ), 1 );
 		add_filter( 'wp_redirect', array( $this, 'log_changes' ) );
 	}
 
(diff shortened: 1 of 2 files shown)


connectors/class-connector-editor.php
```diff
@@ -187,33 +186,34 @@ public function action_links( $links, $record ) {
...
 		}
 
 		$plugin_slug = wp_stream_filter_input( INPUT_POST, 'plugin' );
+
 		if ( $plugin_slug ) {
+			$location          = 'plugin-editor.php';
 			$this->edited_file = $this->get_plugin_data( $plugin_slug );
 		}
+
+		$this->log_changes( $location );
```
tharsheblows, line 216: **❓question**
If there's no `$location` here, should we log anything? I don't think so?

connectors/class-connector-editor.php
```diff
@@ -187,33 +186,34 @@ public function action_links( $links, $record ) {
...
-				isset( $_SERVER['REQUEST_METHOD'] )
-				&&
-				'POST' !== sanitize_text_field( $_SERVER['REQUEST_METHOD'] )
-			)
-			||
-			'update' !== wp_stream_filter_input( INPUT_POST, 'action' )
-		) {
+		$action         = wp_stream_filter_input( INPUT_POST, 'action' );
+		$request_method = wp_stream_filter_input( INPUT_SERVER, 'REQUEST_METHOD' );
+
```
tharsheblows, line 196: **❗️issue**
Can you add a nonce and capability check here? I think it can be the same capability check as is required to do the action.

tharsheblows (changes requested):
A couple of comments but this looks good. I looked it up and the ajax method was added in 4.9 so I think we're safe? It will still work with < WP 4.9 but won't record the changes. 

Should we add a "Requires at least" header? I'm not really sure which version that would be though.

connectors/class-connector-editor.php
marcinkrzeminski: Fixed that.

connectors/class-connector-editor.php
marcinkrzeminski: Fixed that as well.

tharsheblows (approved):
Thank you! It all looks good.

PR #1680, opened 2025-01-15
Fix the build process

bartoszgadomski:
The `build` directory is excluded in the build pipeline, which causes the fatal error:

screen_shot_2025-01-15_at_13 01 46

# Checklist

## Release Changelog

- Fix: Issue with excluding the `build` directory in the build pipeline

### DIFF
--- a/.github/workflows/deploy-to-stream-dist.yml
+++ b/.github/workflows/deploy-to-stream-dist.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - master
       - develop
+  workflow_dispatch:
 
 jobs:
   lint_and_test:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-to-stream-dist.yml
```diff
@@ -57,15 +57,15 @@ jobs:
       - name: Setup environment
         run: |
           set -ex
-          
+
           ROOT_DIR="$(git rev-parse --show-toplevel)"
           WORKING_BRANCH="$(git rev-parse --abbrev-ref HEAD)"
-          SRC_DIR="$ROOT_DIR/build"
+          SRC_DIR="$ROOT_DIR/source"
```
bartoszgadomski, line 63: We can't use the `build` directory, as this one is reserved for the files generated by the `npm run build` command.

PR #1687, opened 2025-01-28
Fix absint Warning

obenland:
Fixes a PHP warning when trying to instantiate `Author` with an `stdClass` object instead of a user ID:

```
Warning: Object of class stdClass could not be converted to int in /wordpress/core/6.7.1/wp-includes/load.php on line 1440
```

Confusingly, `get_users()` works different to `get_posts()`, where asking for only IDs still returns an object with only an ID property instead of an array of IDs.

# Checklist

## Release Changelog

## Release Checklist

### DIFF
--- a/classes/class-list-table.php
+++ b/classes/class-list-table.php
@@ -571,8 +571,8 @@ class List_Table extends \WP_List_Table {
 			}
 
 			$users = array_map(
-				function ( $user_id ) {
-					return new Author( $user_id );
+				function ( $user ) {
+					return new Author( $user->ID );
 				},
 				get_users(
 					array(


dd32:
> Confusingly, `get_users()` works different to `get_posts()`, where asking for only IDs still returns an object with only an ID property instead of an array of IDs.

This PR appears invalid.

`var_dump( get_users( ['fields' => [ 'ID' ] ] ) );` returns an array of user objects with an ID property (As the PR is intending on supporting).

`var_dump( get_users( ['fields' => 'ID' ] ) );` (as in stream) returns an array of IDs.

It appears that something might be altering the `get_users()` return value in your environment @obenland  - although the documentation of this is confusing so I can see why someone would've made an error: https://developer.wordpress.org/reference/functions/get_users/#:~:text=If%20%E2%80%98fields%E2%80%98%20is%20set%20to%20an,an%20array%20of%20IDs%20will%20be%20returned.

obenland:
Oh, that _is_ confusing. Thanks for the heads up, I'll take another look.

PR #1859, opened 2026-05-04
Register Stream operations as WordPress Abilities

PatelUtkarsh:
Resolves [XWPENG-13](https://xwp.atlassian.net/browse/XWPENG-13).

Integrates Stream with the WordPress 6.9 Abilities API. Eleven Stream operations are registered as abilities under the `stream/` namespace, exposed via the core Abilities REST controller at `/wp-abilities/v1/abilities/stream/{slug}/run` once a site owner explicitly opts in.

## Approach

Mirrors Stream's existing connector/alert plugin extension layout: an abstract base class, a loader gated on environment + opt-in setting, and one file per ability that delegates to existing Stream internals (no business logic duplicated).

**Permission model** — read abilities use Stream's `view_stream` capability (via a shared trait) so editors and other roles configured under "Role Access" can call them, matching the admin UI. Write and destructive abilities require the Stream settings capability (`WP_STREAM_SETTINGS_CAPABILITY`, filterable; defaults to `manage_options`).

**Schemas** are hand-written with AI-readable `description` strings rather than auto-derived, decoupling the public API from internal arg names.

**Abilities registered:**

| Ability | Permission | Backed by |
|---|---|---|
| `stream/get-records` | `view_stream` | Stream records query |
| `stream/get-record` | `view_stream` | Single record + meta |
| `stream/get-settings` | settings cap | Stream settings option |
| `stream/get-alerts` | `view_stream` | `wp_stream_alerts` posts + meta |
| `stream/get-connectors` | `view_stream` | Registered connectors + labels |
| `stream/get-exclusion-rules` | `view_stream` | Pivoted exclusion-rule rows |
| `stream/create-alert` | settings cap | New `wp_stream_alerts` post |
| `stream/update-settings` | settings cap | Partial-merge option update |
| `stream/create-exclusion-rule` | settings cap | Append to exclusion-rule columns |
| `stream/purge-records` | settings cap | Filtered cascade delete |
| `stream/delete-alert` | settings cap | Hard-delete `wp_stream_alerts` post |

Routing follows the WP core Abilities REST controller defaults: read-only abilities are served as `GET`, write abilities as `POST`, and destructive + idempotent abilities as `DELETE`. Non-matching methods return `405 Method Not Allowed`.

## Gating

1. **WordPress version** — silent no-op on WP < 6.9. The loader returns early in its constructor, the settings field is hidden in the Advanced section, and the abstract base's `register()` checks `function_exists( 'wp_register_ability' )` defensively. No admin notice is shown (per direction).
2. **Settings toggle** — new `advanced_enable_abilities_api` checkbox in the existing Advanced section of Stream settings. Defaults to `0`. Only renders on WP 6.9+. Honors network option on network-activated multisite.

## Safety

- Read abilities use `view_stream`; write/destructive abilities use the Stream settings capability (filterable).
- `stream/purge-records` requires `confirm: true` AND at least one filter (`older_than_days`, `connector`, `context`, `action`); a confirm-only payload returns a `400` `WP_Error` rather than truncating the table. On any multisite request that is not running inside Network Admin (REST never is), the purge is scoped to `blog_id = get_current_blog_id()` so it cannot wipe other sites' records.
- `stream/get-record` applies the same per-blog scoping so a `view_stream` user on one site can't read records from other sites by guessing IDs.
- `stream/delete-alert` is idempotent — the second call returns the same `404` `WP_Error` as a missing ID. Refuses to delete posts that are not of `wp_stream_alerts` type.
- `stream/create-alert` validates `alert_type` against registered alert types and rejects unknown types.
- `stream/get-records` schema tightened: `orderby` is an enum, `*__in` arrays have `maxItems` caps.
- `stream/create-exclusion-rule` validates IP format at the application layer.
- `older_than_days` cutoffs are computed as a UTC `DateTime` in PHP (matching `Admin::purge_scheduled_action()`) so behavior is independent of MySQL server timezone.

## Manual REST QA

After merging and enabling the toggle on a WP 6.9+ site:

```bash
# List the eleven abilities.
curl -u admin:password 'http://stream.wpenv.net/wp-json/wp-abilities/v1/abilities?category=stream'

# Read connectors (works for any role with view_stream). Read abilities are GET.
curl -u editor:password \
  'http://stream.wpenv.net/wp-json/wp-abilities/v1/abilities/stream/get-connectors/run'

# Create an alert (requires settings cap). Write abilities are POST.
curl -u admin:password -X POST \
  'http://stream.wpenv.net/wp-json/wp-abilities/v1/abilities/stream/create-alert/run' \
  -H 'Content-Type: application/json' \
  -d '{"input":{"alert_type":"highlight","trigger_author":"any","trigger_context":"any","trigger_action":"any"}}'

# Filtered purge. Destructive + idempotent abilities are DELETE.
curl -u admin:password -X DELETE \
  'http://stream.wpenv.net/wp-json/wp-abilities/v1/abilities/stream/purge-records/run' \
  -H 'Content-Type: application/json' \
  -d '{"input":{"confirm":true,"older_than_days":90}}'

# Unauthenticated request — expect 401/403.
curl 'http://stream.wpenv.net/wp-json/wp-abilities/v1/abilities/stream/get-records/run'
```

# Checklist

## Release Changelog

- New: Register 11 Stream operations as WordPress 6.9 Abilities API abilities under the `stream/` namespace, gated behind a new "Enable Abilities API" advanced setting (default off). Read abilities respect Stream's `view_stream` capability; write abilities require the Stream settings capability.

[XWPENG-13]: https://xwp-co.atlassian.net/browse/XWPENG-13?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


abilities/class-ability-purge-records.php
```diff
@@ -0,0 +1,176 @@
...
+				__( 'Purge requires confirm: true.', 'stream' ),
+				array( 'status' => 400 )
+			);
+		}
+
+		$where  = array( '1=1' );
+		$params = array();
+
+		if ( ! empty( $input['older_than_days'] ) ) {
+			$where[]  = 'stream.created < DATE_SUB(NOW(), INTERVAL %d DAY)';
```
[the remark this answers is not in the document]
PatelUtkarsh: Good catch!

abilities/class-ability-create-alert.php
```diff
@@ -0,0 +1,195 @@
...
+		update_post_meta( $post_id, 'alert_type', $input['alert_type'] );
+		update_post_meta( $post_id, 'alert_meta', $alert_meta );
+
+		return array(
+			'id'         => (int) $post_id,
+			'status'     => (string) get_post_status( $post_id ),
+			'title'      => (string) get_the_title( $post_id ),
+			'alert_type' => get_post_meta( $post_id, 'alert_type', true ),
+			'alert_meta' => (array) get_post_meta( $post_id, 'alert_meta', true ),
+		);
```
bartoszgadomski: @PatelUtkarsh Instead of duplicating the logic in this class and method, could you call the `save` method on the Alert class? Ideally, abilities should use existing data flows to keep things consistent and properly separated.

https://github.com/xwp/stream/blob/477d9f892f6907abd7c88d9c7b119e703326a0ea/classes/class-alert.php#L86-L92

abilities/class-ability-create-alert.php
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	/**
+	 * {@inheritDoc}
+	 *
+	 * @param mixed $input Validated input matching get_input_schema(), or null.
+	 */
+	public function execute( $input = null ) {
+		$status = isset( $input['status'] ) ? $input['status'] : 'wp_stream_enabled';
```
bartoszgadomski: @PatelUtkarsh Should the value of `$input['status']` be validated against an enum, or is that ensured by the API?

abilities/class-ability-delete-alert.php
```diff
@@ -0,0 +1,104 @@
...
+
+		if ( ! $post || Alerts::POST_TYPE !== $post->post_type ) {
+			return new \WP_Error(
+				'stream_alert_not_found',
+				__( 'Alert not found.', 'stream' ),
+				array( 'status' => 404 )
+			);
+		}
+
+		$result = wp_delete_post( $id, true );
```
bartoszgadomski: @PatelUtkarsh Similar concern here; while Alert class currently has no `delete` method, I think we should add it there and let this ability to call it rather than executing these operations directly.

abilities/class-ability-get-alerts.php
```diff
@@ -0,0 +1,145 @@
...
+
+		switch ( $requested ) {
+			case 'enabled':
+				$statuses = array( 'wp_stream_enabled' );
+				break;
+			case 'disabled':
+				$statuses = array( 'wp_stream_disabled' );
+				break;
+			default:
+				$statuses = array( 'wp_stream_enabled', 'wp_stream_disabled' );
```
bartoszgadomski: @PatelUtkarsh Should these `wp_stream_enabled` and `wp_stream_disabled` magic strings be moved to const in Ability class instead? I can see them repeated multiple times across abilities.

abilities/class-ability-get-alerts.php
```diff
@@ -0,0 +1,145 @@
...
+			}
+
+			$out[] = array(
+				'id'         => (int) $post->ID,
+				'status'     => (string) $post->post_status,
+				'title'      => (string) $post->post_title,
+				'alert_type' => get_post_meta( $post->ID, 'alert_type', true ),
+				'alert_meta' => $alert_meta,
+			);
+		}
```
bartoszgadomski: Same concern here re: directly executing logic that belongs to another class (Alerts in this case).

abilities/class-ability-get-connectors.php
```diff
@@ -0,0 +1,114 @@
...
+		$connectors = isset( $this->plugin->connectors->connectors ) ? (array) $this->plugin->connectors->connectors : array();
+
+		foreach ( $connectors as $slug => $connector ) {
+			$out[] = array(
+				'slug'     => (string) $slug,
+				'label'    => method_exists( $connector, 'get_label' ) ? (string) $connector->get_label() : (string) $slug,
+				'contexts' => method_exists( $connector, 'get_context_labels' ) ? (array) $connector->get_context_labels() : array(),
+				'actions'  => method_exists( $connector, 'get_action_labels' ) ? (array) $connector->get_action_labels() : array(),
+			);
+		}
```
bartoszgadomski: @PatelUtkarsh Could this be moved to Connectors class instead, under `get_all` method or similar?

abilities/class-ability-get-record.php
```diff
@@ -0,0 +1,143 @@
...
+
+		if ( empty( $row ) ) {
+			return new \WP_Error(
+				'stream_record_not_found',
+				__( 'Record not found.', 'stream' ),
+				array( 'status' => 404 )
+			);
+		}
+
+		$row['meta'] = (array) get_metadata( 'record', $row['ID'] );
```
bartoszgadomski: @PatelUtkarsh Another case, and I think I'll stop here :P IMO this should be added as a new method under Record class and simply called here. You can make shape adjustments to fit the output into the ability schema, but direct logic should live in dedicated classes instead.

classes/class-abilities.php
```diff
@@ -0,0 +1,277 @@
...
+				array()
+			);
+		} else {
+			$options = isset( $this->plugin->settings )
+				? (array) $this->plugin->settings->options
+				: array();
+		}
+
+		return ! empty( $options[ $key ] );
+	}
```
bartoszgadomski, line 194: @PatelUtkarsh Is there no existing way to get a single setting value (toggle in this case)? Calling this directly feels too heavy; should we have a `get_setting_value` method in Settings class instead and let it run this logic internally?

bartoszgadomski (changes requested):
@PatelUtkarsh Thank you for working on this exciting new feature!

Please see my feedback below - it's mostly related to separation of concerns and de-duplicating logic.

abilities/class-ability-create-alert.php
PatelUtkarsh: Good call! Fixed in 8e4637a. `execute()` builds an `Alert` and calls `save()` now.

abilities/class-ability-create-alert.php
PatelUtkarsh: The schema declares the enum in `get_input_schema()` and the Abilities API validates against it before `execute()` runs, so invalid values get rejected with a 400 at the API boundary. Left as-is.

abilities/class-ability-delete-alert.php
PatelUtkarsh: Good call! Fixed in 8e4637a. Added `Alert::delete()` (checks post_type, then `wp_delete_post`). Ability now does `Alerts::get_alert()` + `$alert->delete()`.

abilities/class-ability-get-alerts.php
PatelUtkarsh: Good call! Fixed in 8e4637a. Added `Alerts::STATUS_ENABLED` and `Alerts::STATUS_DISABLED`. Put them on `Alerts` since they describe alert post statuses.

abilities/class-ability-get-alerts.php
PatelUtkarsh: Good call! Fixed in 8e4637a. Added `Alerts::get_alerts( array $statuses )`. Ability just calls it and shapes the response.

abilities/class-ability-get-connectors.php
PatelUtkarsh: Good call! Fixed in 8e4637a. Added `Connectors::get_all()` and `Connectors::get_slugs()`. Both abilities delegate.

abilities/class-ability-get-record.php
PatelUtkarsh: Good call! Fixed in 8e4637a. Added static `Record::get_by_id( $id, $blog_id = null )`. The ability passes the current blog id on multisite outside network admin, so the model stays neutral.

classes/class-abilities.php
PatelUtkarsh: Good call! Fixed in 8e4637a. Added `Settings::get_setting_value( $key, $default )` that handles the network-activated case internally. `is_enabled()` is now a one-liner.

PatelUtkarsh:
Thanks for the thorough review! @bartoszgadomski Addressed all the feedback in [8e4637a](https://github.com/xwp/stream/pull/1859/commits/8e4637aad3d18ab0daa4d6e324bc001e3461216e).

The main theme was routing abilities through existing classes `(Alert::save(), Alert::delete(), Alerts::get_alerts(), Connectors::get_all(), Record::get_by_id(), Settings::get_setting_value())` instead of duplicating logic in the ability layer.

abilities/class-ability-get-records.php
```diff
@@ -0,0 +1,261 @@
...
+			'orderby',
+		);
+
+		$args = array_intersect_key( $input, array_flip( $allowed_keys ) );
+
+		if ( ! isset( $args['records_per_page'] ) ) {
+			$args['records_per_page'] = self::DEFAULT_PER_PAGE;
+		}
+
+		$records = $this->plugin->db->get_records( $args );
```
bartoszgadomski, line 254: @PatelUtkarsh `get-record` and `purge-records` were given an explicit `is_multisite() && ! is_network_admin()` blog guard, but `get-records` was missed. The reasoning that protects the others applies here too, and the `wp_stream_query_args` filter that normally injects `blog_id` is never registered. A user with `view_stream` on any blog can call `stream/get-records` and receive records from every other blog.

Apply the same scoping as `get-record` / `purge-records`:

```suggestion
		if ( is_multisite() && ! is_network_admin() ) {
			$args['blog_id'] = get_current_blog_id();
		}

		$records = $this->plugin->db->get_records( $args );
```

abilities/class-ability-update-settings.php
```diff
@@ -0,0 +1,146 @@
...
+			}
+		}
+
+		// Run only the incoming keys through Stream's sanitize pipeline so
+		// values are normalized to their declared field type, then merge over
+		// the existing options so unrelated keys are preserved.
+		$sanitized = $this->plugin->settings->sanitize_settings( $filtered );
+		$merged    = array_merge( $current, $sanitized );
+
+		update_option( $option_key, $merged );
```
bartoszgadomski: @PatelUtkarsh `is_enabled()` method was fixed to read via `Settings::get_setting_value()` so it consults `wp_stream_network` on network-activated installs. The writes weren't; In REST on a network-activated install, `Settings::get_option_key()` returns `wp_stream` (the per-site key), but the authoritative store is `wp_stream_network` (written by `Network::network_admin_save_options()` via `update_site_option`). 

Result: an admin who flips a setting via the API thinks it was saved, but `Abilities::is_enabled()`, `Admin::purge_scheduled_action()`, and the in-network-admin UI all still read the old value. This should route through a setter on Settings that knows about the network case.

abilities/class-ability-create-exclusion-rule.php
```diff
@@ -0,0 +1,228 @@
...
+
+		$rule = array();
+		foreach ( self::RULE_COLUMNS as $column ) {
+			$value                      = $sanitized[ $column ];
+			$rules[ $column ][ $index ] = $value;
+			$rule[ $column ]            = '' === $value ? null : $value;
+		}
+
+		$options['exclude_rules'] = $rules;
+		update_option( $option_key, $options );
```
bartoszgadomski: Same concern applies here.

classes/class-abilities.php
```diff
@@ -0,0 +1,262 @@
...
+	 * @return array
+	 */
+	public function filter_user_caps( $allcaps, $caps, $args, $user = null ) {
+		unset( $args );
+
+		if ( ! in_array( 'view_stream', (array) $caps, true ) ) {
+			return $allcaps;
+		}
+
+		$role_access = isset( $this->plugin->settings->options['general_role_access'] )
```
bartoszgadomski: Same multisite gap on the read side: the role-access dynamic cap filter reads `$plugin->settings->options['general_role_access']`. On network-activated multisite REST contexts, `$plugin->settings->options` is the (typically empty) per-site option. So users who would be granted `view_stream` via the network admin's Role Access setting won't get it through the abilities endpoint - read abilities silently return 403 for any non-administrator. Use `Settings::get_setting_value( 'general_role_access', array() )` here, mirroring what was done for the toggle.

classes/class-record.php
```diff
@@ -135,6 +135,51 @@ public function __construct( $item ) {
...
+		// user input), and $prepared holds only integer IDs.
+		$sql = "SELECT * FROM {$wpdb->stream} WHERE ID = %d{$where}";
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery,WordPress.DB.DirectDatabaseQuery.NoCaching,WordPress.DB.PreparedSQL.NotPrepared
+		$row = $wpdb->get_row( $wpdb->prepare( $sql, $prepared ), ARRAY_A );
+
+		if ( empty( $row ) ) {
+			return null;
+		}
+
+		$row['meta'] = (array) get_metadata( 'record', $row['ID'] );
```
bartoszgadomski, line 178: When the record has no meta, `get_metadata()` returns `[]`, which JSON-encodes as `[]`, violating the declared `meta: { type: object }` in the output schema. The same 5117ef1 commit that fixed `alert_meta` should be applied here - coerce empty to `new \stdClass()` (or `is_array() && ! empty()` ternary).

bartoszgadomski (changes requested):
@PatelUtkarsh Thank you for implementing the requested changes!

I have a few more findings - note these are AI generated, so please double-check whether there's no any false alarms. Beside the inline comments shared below for critical issues, here are other findings:

## Medium

### `advanced_enable_abilities_api` is unreachable through `stream/update-settings` on network-activated multisite
`Settings::get_fields()` only appends the toggle when `! is_network_activated() || is_network_admin()`. REST is never `is_network_admin()`, so on a network-activated install the field is missing from `get_fields()` everywhere except the network-admin settings screen. `update-settings` builds its allowlist from `get_fields()`, so the key gets `array_intersect_key`-stripped and the call fails with `stream_no_valid_settings`. This is arguably intentional (force the toggle through the network admin UI), but it should be either documented or — better — whitelisted explicitly so an MCP client can still flip it. Same gap applies to `get-settings`, where the toggle's current value will never appear in the response on network-activated installs.

### `stream/get-connectors` returns a frontend-only subset, breaking `stream/create-exclusion-rule` validation
`Connectors::load_connectors()` honors `register_frontend = false` and skips those connectors when `! is_admin()` — REST satisfies `! is_admin()`. The following connectors set `register_frontend = false`: `blogs`, `taxonomies`, `settings`, `editor`, `menus`, `installer`, `jetpack`, `mercator`. Two visible consequences:

1. `get-connectors` returns a list that's a *strict subset* of the connectors an admin sees in the UI, which is misleading for AI clients trying to enumerate what's tracked on the site.
2. `create-exclusion-rule` validates `connector` against `$plugin->connectors->get_slugs()`. Trying to exclude `connector: "settings"` or `connector: "editor"` from the REST API will hit `stream_unknown_connector`, even though those connectors are real, log records, and are valid in the UI.

The cleanest fix is to either force `register_frontend = true` for connectors that should be discoverable via abilities, or to drive the validation list from a non-context-dependent source.

### `update-settings` re-triggers the TTL purge hook
`Settings::__construct()` hooks `update_option_$option_key` → `updated_option_ttl_remove_records`. `update-settings` writes through `update_option`, so any REST call that changes `general_records_ttl` (or even just touches the option) will fire that hook in the REST request. Likely desired, but worth confirming — historically that hook scheduled an async cascade purge, and it now runs in user-facing REST latency. Either reuse the existing path explicitly or document.

## Lower priority / nits

- **`Ability_Get_Alerts::execute()`** casts `(string) $alert->status` and returns it. The schema enums to `[STATUS_ENABLED, STATUS_DISABLED]`, but `Alert::__construct` actually keeps whatever the post status is (e.g. `trash`). Since the listing only fetches `wp_stream_enabled`/`wp_stream_disabled` from `Alerts::get_alerts()`, this is fine *today* — but if a third-party plugin trashes an alert and you later expand `get_alerts()` to include other statuses, the output will violate the enum. Cheap defensive coercion or an `if ( ! in_array( $status, ..., true ) ) continue;` would future-proof this.

- **`Ability_Create_Alert` output schema vs. reality**: the response builds `alert_type` via `get_post_meta(...)` which returns `''` (not `null`) when meta is missing. Schema is `type: [string, null]`, so an empty string is fine — but `alert_meta` is cast `(array) get_post_meta(...)` and if `Alert::save()` ever leaves meta unsaved (e.g. a failed `update_post_meta`), `(array) ''` becomes `[ 0 => '' ]`, a list. Same JSON-shape bug class as the `get-record` meta shape. Currently unreachable because `save()` always writes a non-empty `alert_meta`, but worth a defensive normalize before returning.

- **`Trait_View_Stream_Permission` lives in `abilities/trait-view-stream-permission.php` and is `require_once`'d at the top of every ability file that uses it.** With the loader's `include_once` and PHPUnit's per-test `require_once`, this works, but it's also fragile: rename the file or change the autoload prefix and the abilities silently lose their permission callback (and fall back to the `WP_STREAM_SETTINGS_CAPABILITY` default — which would *over*-restrict reads, not under-restrict). Consider moving the trait to `classes/` so the existing PSR-4-ish autoloader picks it up via `Trait_View_Stream_Permission` → `class-trait-view-stream-permission.php`, or rename to live under the existing autoload conventions.

- **`Abilities::register_category` / `register_abilities` / `filter_user_caps` callbacks** are registered as `array( $this, '...' )`. The workspace coding rule (`Avoid Anonymous Functions in PHP`) discourages instance-method callbacks in favor of static or named functions, but the rest of Stream consistently uses `array( $this, '...' )` (e.g. `Settings::__construct()`, `Admin::__construct()`). New code is therefore consistent with the repo but inconsistent with the stated standard. Worth a project-level decision before this lands; not a blocker.

- **`Test_Ability_Update_Settings::test_partial_update_preserves_other_keys`** and friends pass only because they exercise the single-site code path. Once the multisite write issue is addressed, you'll want a `@group ms-required` test that proves a write lands in `wp_stream_network` on network-activated installs (mirroring `test_is_enabled_reads_network_option_when_network_activated`).

- **`Test_Abilities::test_load_abilities_instantiates_each_slug`** and `test_load_abilities_populates_all_slugs` are functionally duplicates — both assert count 11 + iterate the slugs. Worth collapsing.

- **`Ability_Get_Settings`** doesn't override the base permission_callback, so it uses `WP_STREAM_SETTINGS_CAPABILITY`. The PR description says read abilities use `view_stream` and only write/destructive abilities use the settings cap; the doc is slightly off — `get-settings` is read but gated on the settings cap. Reasonable choice (settings can include role allowlists), but worth documenting why this one read ability is treated differently.

- **`Ability::register()` includes a `final` qualifier**, which is great for the contract, but it means a subclass that ever needed to inject e.g. a precondition (`register_only_if_woocommerce_loaded`) has nowhere to hook. Add a `pre_register()` hook method or remove `final` later if needed; minor.

abilities/class-ability-get-records.php
PatelUtkarsh: Good catch! Fixed in 71d64433. Added the same `is_multisite() && ! is_network_admin()` guard `get-record` and `purge-records` have.

abilities/class-ability-update-settings.php
PatelUtkarsh: Good catch! Fixed in 71d64433. Added `Settings::update_all_setting_values()` and `get_all_setting_values()` that route through `update_site_option`/`get_site_option` on network-activated multisite.

abilities/class-ability-create-exclusion-rule.php
PatelUtkarsh: Fixed in 71d64433 alongside `update-settings`.

classes/class-abilities.php
PatelUtkarsh: Good catch! Fixed in 71d64433. `filter_user_caps()` now reads via `Settings::get_setting_value()`.

classes/class-record.php
PatelUtkarsh: Fixed in 71d64433. Coerced empty `meta` to `new \stdClass()` in `Ability_Get_Record::execute()`. Keeps JSON-shape concerns in the ability layer and the model neutral.

PatelUtkarsh:
Thanks for the second pass! All five inline points addressed in 71d64433. @bartoszgadomski 

**Medium**
- **`advanced_enable_abilities_api` unreachable on network-activated REST**: fixed in 71d64433.
- **`get-connectors` frontend-only subset / `create-exclusion-rule` rejects valid connectors**: fixed in 0f8e5a0b. Added `Connectors::get_all_including_admin_only()` that ignores the `register_frontend` gate, metadata-only, never calls `register()`.
- **TTL purge hook on REST writes**: same hook fires when admin UI saves. No code change.

**Nits addressed in d1fc2e07:**
- Defensive status enum filter in `Ability_Get_Alerts`.
- Defensive empty-`alert_meta` coerce in `Ability_Create_Alert`.
- New `Test_Ability_Update_Settings::test_write_targets_network_option_when_network_activated`.
- Merged duplicate `Test_Abilities` slug tests.
- Documented `Ability_Get_Settings` capability choice.

**`Trait_View_Stream_Permission` fragility**: fixed in a65b9cfa. Centralized the `require_once` in `Abilities::load_abilities()` and the shared test case. New read abilities pick up the trait automatically.

Deferring:
- `array( $this, '...' )` callbacks: consistent with rest of Stream; repo-wide decision.
- `final` on `Ability::register()`: YAGNI.

bartoszgadomski:
@PatelUtkarsh thank you for your excellent work on this! I tested this locally and have a few observations:

1. There are a few prerequisites that aren't met by default in this repo for local testing of this feature. Could we add them in this PR so that those who want to test locally don't have to figure them out themselves?

    - HTTPS, to enable application passwords
    - The [wordpress/mcp-adapter](https://github.com/WordPress/mcp-adapter) plugin as a dependency in composer.json

2. Even after I manually addressed the above, enabled the Abilities API toggle in the Stream plugin settings, loaded some test records, configured an MCP server, and connected to it through Claude, the Stream plugin returns an empty list of abilities. I also installed and activated the [Abilities API plugin](https://github.com/WordPress/abilities-api) in case it was needed, but the list is still empty. Am I missing a step? Note that I did restart Claude every time the configuration changed (and I'm on the branch ;)

---

Screenshot 2026-05-19 at 11 07 25

.github/workflows/lint-and-test.yml
```diff
@@ -134,6 +134,16 @@ jobs:
...
 
+      - name: Rebuild WordPress image on top of the pulled base
+        # The published ghcr.io/xwp/stream-wordpress image is updated only by
+        # docker-images.yml on master pushes. When this branch changes
+        # local/docker/wordpress/Dockerfile (e.g. to enable mod_ssl), the
+        # pulled image lags behind the branch's intent and Apache rejects
+        # :443 with ERR_CONNECTION_CLOSED. Build locally so the branch's
+        # Dockerfile actually applies; Docker layer caching keeps this cheap
+        # when the Dockerfile hasn't changed since the last published build.
+        run: docker compose build wordpress
```
PatelUtkarsh, line 145: Note this needs to be removed once PR is merged to master.

PatelUtkarsh:
@bartoszgadomski You should able to use this as MCP now. More details here: https://github.com/xwp/stream/blob/feature/XWPENG-13-abilities-api/contributing.md#mcp-model-context-protocol-integration

bartoszgadomski (approved):
I tested this locally, works well. Excellent work @PatelUtkarsh, thank you!

PR #1915, opened 2026-07-02
Fix Heartbeat authorization bypass (CVE-2026-11907) and related hardening

bartoszgadomski:
## Summary

Fixes the missing-authorization vulnerability reported as **CVE-2026-11907** (Wordfence), plus additional authorization/hardening issues found during a follow-up review of the same code class.

### CVE-2026-11907 — Heartbeat API sensitive information disclosure
`Live_Update::heartbeat_received()` is hooked into the global `heartbeat_received` filter, which fires for any authenticated user on any admin screen. It built the Stream `List_Table` and returned rendered activity rows without checking that the user is allowed to view Stream records, letting Subscriber-level users read the full activity log via `admin-ajax.php`.

- Add a `view_stream` capability check before any data is prepared.
- Pass the `$single` argument to `get_user_meta()` so the live-update preference is read as a scalar instead of always evaluating truthy.

### Additional hardening (found during review)
- **`Alerts::get_actions()`** (`wp_ajax_get_actions`): previously had no authorization check and was callable by any authenticated user. Added a capability check consistent with the other alert AJAX handlers.
- **`Alert_Type_Highlight::ajax_remove_highlight()`**: added a `view_stream` capability check alongside the existing nonce (defense-in-depth).
- **`Export::render_download()`**: require the `view_stream` capability before streaming records instead of relying on the nonce alone.
- **`Query::query()`**: restrict the selectable `fields` to an allowlist (column names cannot be passed through `$wpdb->prepare()`), closing a latent SQL injection sink; fall back to `SELECT *` when no valid field is supplied.
- **`CLI::query()`**: pass `fields` as an array so each column is validated individually by the query builder, preserving exact column selection with the new allowlist.

## Test plan
- [ ] As a Subscriber, confirm the Heartbeat PoC (`data[wp-stream-heartbeat]=live-update`) no longer returns `wp-stream-heartbeat` rows or `total_items`.
- [ ] As an Administrator on the Stream Records page, confirm live updates still work.
- [ ] Confirm `wp_ajax_get_actions` returns `-`/error for users without the settings capability, and still works in the New Alert form for admins.
- [ ] Confirm highlight removal and CSV/JSON export still work for authorized users and are rejected for unauthorized ones.
- [ ] `wp stream query --fields=created,ip,summary --format=json` returns only the requested columns.
- [ ] Run the PHPUnit suite.

Made with [Cursor](https://cursor.com)


classes/class-alerts.php
```diff
@@ -774,6 +774,14 @@ public function get_notification_values() {
 	 * Update actions dropdown options based on the connector selected.
 	 */
 	public function get_actions() {
+		if ( ! current_user_can( self::CAPABILITY ) ) {
```
PatelUtkarsh, line 778: **suggestion (non-blocking):** `get_actions()` now has a capability check but still no nonce, unlike `save_new_alert()` and `load_alerts_settings()` in this same class. It is read-only so the risk is low, but it leaks connector action labels to any same-origin request and breaks the plugin's own convention. Consider adding `check_ajax_referer()` here too.

classes/class-query.php
```diff
@@ -206,16 +206,27 @@ public function query( $args ) {
 		$fields  = (array) $args['fields'];
 		$selects = array();
 
+		// Column names cannot be passed through $wpdb->prepare(), so restrict
+		// the selectable fields to a known allowlist to prevent SQL injection
+		// via the `fields` argument.
+		$selectable_fields = array( 'ID', 'site_id', 'blog_id', 'object_id', 'user_id', 'user_role', 'created', 'summary', 'connector', 'context', 'action', 'ip' );
```
PatelUtkarsh: **nitpick:** This allowlist duplicates `$allowed_fields` at line 61 verbatim. Extract a class constant (e.g. `Query::ALLOWED_FIELDS`) so the two lists cannot drift.

PatelUtkarsh (approved):
Test is failing but otherwise PR looks good.

classes/class-alerts.php
bartoszgadomski: Done, added `check_ajax_referer()` and passed the nonce from both callers.

classes/class-query.php
bartoszgadomski: Done, extracted the shared allowlist to `Query::ALLOWED_FIELDS`.

PR #1917, opened 2026-07-03
Make Action Scheduler usage optional at runtime (WP-Cron fallback)

### LINKED ISSUE #1907: Reconsider need for Action Scheduler dependency
With Stream 4.1+, Action Scheduler is now a dependency of Stream. This seems to have been introduced in #1543 as a way to deal with large tables in cases where cron may be unreliable.

However, this is a pretty hefty requirement to add in for a less common case, and in some cases it may not actually be necessary. When using a third-party database driver (i.e. implementing the `DB_Driver` interface), purging old records may not be desirable or necessary; on smaller sites, the overhead of AS for processing large record counts isn't necessary anyway. Additionally, when using a service like Cavalcade for reliable cron, wp-cron tasks work fine anyway.

This is an issue for us on Altis as Action Scheduler operates less efficiently for this sort of processing than wp-cron, and we bundle Stream. We also use a custom storage driver as we store records in a separate tamper-proof database.

I'd like to propose moving AS to being an optional dependency of Stream. This would entail two changes:
* Switch the action scheduling functions to _conditionally_ call AS - if it's loaded, use `as_enqueue_async_action()` etc, otherwise use regular wp-cron
* Move AS from being a dependency to being a `suggest` package

This would allow using AS on the sites that need it, while not needing it for the common case and when not necessary.

PatelUtkarsh:
Fixes #1907.

## Why

Stream 4.2.0 made Action Scheduler a hard dependency for deferred purge / reset work. That's a problem for hosts that bundle Stream and already run reliable cron (e.g. Altis / Cavalcade) — they don't want AS's tables, its `init` bootstrap, or a second queue system.

## What

- New `Scheduler` interface with two backends: `AS_Scheduler` (default, unchanged behavior) and `Cron_Scheduler` (WP-Cron fallback).
- Select the fallback with `add_filter( 'wp_stream_use_action_scheduler', '__return_false' )` — must be registered before Stream loads (mu-plugin). When selected, the AS library is not loaded at all.
- Switching backends needs no migration: a backend marker option clears the stale recurring purge action on the next request, so only one scheduler ever fires.
- New `wp_stream_enable_auto_purge` filter (default `true`) to disable TTL purge scheduling entirely, for storage drivers that manage retention externally. Honored on both scheduling and executing paths.
- On the cron backend, a large-table purge/reset surfaces an admin notice / `WP_CLI::warning` recommending `wp cron event run --due-now`, since traffic-triggered WP-Cron can stall long batch chains.

Action Scheduler remains a bundled dependency for the WordPress.org build.

## Testing

New suites cover backend selection, backend handoff (both directions), the cron scheduler primitives, and the full purge workflow (batch chaining, reaper, overlap guard, master switch, notice matrix) on the cron backend. Existing `Test_Admin` suite continues to cover AS. PHPCS clean; PHPUnit passing on single-site and multisite.

# Checklist

## Release Changelog

- New: Action Scheduler usage is now optional at runtime via the `wp_stream_use_action_scheduler` filter, with a WP-Cron fallback; switching backends needs no migration (#1907).
- New: `wp_stream_enable_auto_purge` filter (default `true`) to disable TTL auto-purge scheduling for storage drivers that manage retention externally (#1907).
- Fix: Guard the Action Scheduler `require_once` so environments that omit or provide their own copy no longer fatal on load (#1907).


classes/class-admin.php
```diff
@@ -757,7 +769,62 @@ private function schedule_erase_large_records( int $log_size ) {
...
+	private function maybe_warn_large_table_without_action_scheduler( int $record_count, string $operation ) {
+		if ( $this->plugin->scheduler instanceof AS_Scheduler ) {
+			return;
+		}
+
+		if ( ! $this->plugin->is_large_records_table( $record_count ) ) {
+			return;
+		}
+
+		$this->notice(
```
bartoszgadomski: This notice is queued via `Admin::notice()`, which (outside WP-CLI) only appends to `$this->notices` and echoes on the `shutdown` hook of the *current* request. Neither call site is a rendered admin page, so the message never reaches an operator in the dashboard:

- **Manual reset:** `wp_ajax_reset()` calls this and then `wp_safe_redirect()` + `exit`. The reset control is a plain anchor to `admin-ajax.php` (guarded only by a JS `confirm()` in `admin.js`), so the browser follows the 302 and discards the response body where the notice was echoed. The landing page shows only the generic `message=data_erased`.
- **Auto-purge:** `purge_scheduled_action()` runs as the recurring callback under `DOING_CRON` on the cron backend, so the notice is echoed into the discarded cron response. It only surfaces under WP-CLI (`wp cron event run`) — the very command it recommends.

The new cron-purge tests assert on the in-memory `$this->admin->notices` array, not on rendering, so they stay green while the notice is effectively invisible.

Net effect: the AC "Manual 'Reset Stream Database' on large tables shows a clear notice when AS is unavailable" isn't met in the browser, and a low-traffic default-cron site can stall a large chain with no visible warning. Suggest persisting the warning (transient / autoloaded option) and rendering it on the next admin page load, or passing a dedicated `message` query arg on the reset redirect.

classes/class-admin.php
```diff
@@ -757,7 +769,62 @@ private function schedule_erase_large_records( int $log_size ) {
 			'blog_id'    => (int) get_current_blog_id(),
 		);
 
-		as_enqueue_async_action( self::ASYNC_DELETION_ACTION, $args );
+		$this->plugin->scheduler->enqueue_async( self::ASYNC_DELETION_ACTION, $args );
```
bartoszgadomski, line 796: Consistency: the manual-reset chain (`schedule_erase_large_records()` → `erase_large_records()`) doesn't call `mark_running()` / `mark_done()`, unlike the auto-purge chain. Under the cron backend, `is_running_async_deletion()` therefore depends solely on the pending cron event; during the callback's execution window (the event is dequeued before it runs, and the next one isn't enqueued yet) it momentarily reports idle, briefly re-exposing the reset link in Settings. Low risk given the background context, but worth either applying the same running-marker treatment used by `auto_purge_batch()` / `auto_purge_reaper()`, or adding a note documenting the intentional asymmetry.

classes/class-plugin.php
```diff
@@ -205,6 +243,53 @@ public function __construct() {
...
+
+		// Load the bundled AS library before instantiating its scheduler.
+		// file_exists() guards an environment that omits AS; AS's own
+		// ActionScheduler_Versions arbitration handles a host that already
+		// provides its own copy.
+		if ( $this->action_scheduler_available ) {
+			require_once $this->locations['dir'] . '/vendor/woocommerce/action-scheduler/action-scheduler.php';
+		}
+
+		return new AS_Scheduler();
```
bartoszgadomski, line 300: Edge case: if an integrator forces `wp_stream_use_action_scheduler` to `true` while the bundled AS library is absent (`$this->action_scheduler_available === false`), this returns an `AS_Scheduler` without loading AS. `enqueue_async()` / `schedule_recurring()` / `next_scheduled()` / `has_scheduled()` then call `as_*()` unguarded and fatal. Only reachable via an explicit `__return_true` override with AS missing (the default path is safe, since the filter defaults to `$this->action_scheduler_available`), but a guard would be more robust — e.g. fall back to `Cron_Scheduler` when `! $this->action_scheduler_available`, or `function_exists()`-guard the remaining `AS_Scheduler` methods the way the two probe methods already are.

bartoszgadomski (commented):
@PatelUtkarsh Thank you for working on this PR!

Please check inline comments below:

classes/class-plugin.php
PatelUtkarsh: Fixed in 5c27ffb1 using the fallback approach: `create_scheduler()` now returns a `Cron_Scheduler` when the filter forces true but `$this->action_scheduler_available` is false, rather than handing back an `AS_Scheduler` whose unguarded `as_*()` calls would fatal.

classes/class-admin.php
PatelUtkarsh: Good catch! Fixed in 5c27ffb1 by persisting the warning instead of echoing it. Outside `WP-CLI`, `maybe_warn_large_table_without_action_scheduler()` now writes the message to a non-autoloaded option (`wp_stream_large_table_cron_notice`), and a new `display_large_table_cron_notice()` on `admin_notices` / `network_admin_notices` renders it on the next admin page load for users with the Stream settings capability, then deletes it. WP-CLI keeps the immediate `WP_CLI::warning`.

classes/class-admin.php
PatelUtkarsh: The asymmetry wasn't intentional; applied the same running-marker treatment in 5c27ffb1. `erase_large_records()` now calls `mark_running( 'async_deletion' )` on entry and `mark_done()` on the terminal batch (no reaper handoff on this chain, so it can clear immediately), and `is_running_async_deletion()` routes through `any_pending_or_running()` instead of the pending-only probe, closing the execution-window gap where the reset link briefly reappeared.

bartoszgadomski (approved):
Thanks @PatelUtkarsh — commit 5c27ffb1 resolves all three points from my earlier review.

One optional, non-blocking nit (also raised by Copilot): `wp_stream_use_action_scheduler` isn't cast to `bool` in `create_scheduler()`, so a filter returning a truthy string (e.g. `'false'`) would select AS. The documented `__return_false` / `__return_true` usage returns real booleans, so this is robustness rather than a real bug — a `(bool)` cast would make selection deterministic.

Approving, nice work!

PR #1918, opened 2026-07-06
Harden live update preference authorization

bartoszgadomski:
## Summary

Security hardening for the `stream_enable_live_update` AJAX handler that toggles the per-user "Live updates" screen option.

- Enforce the Stream view capability (`view_cap`) before processing the request.
- Always target `get_current_user_id()` instead of a client-supplied `user` value, so a user can only ever change their own live update preference.
- Remove the now-unused user field from the screen-option markup (`classes/class-list-table.php`) and the AJAX payload (`src/js/admin.js`).
- Switch the handler to `wp_stream_filter_input()` for the `checked`/`heartbeat` params for consistent request handling.

## Tests

Adds `tests/phpunit/test-class-live-update.php`:

- `test_enable_live_update_only_updates_current_user()` — a user with view access cannot modify another user's preference by supplying a forged `user` id.
- `test_enable_live_update_denies_users_without_view_cap()` — a user without view access is denied.

## Notes

- No behavior change for legitimate use: an authorized user still toggles their own preference from the Screen Options panel exactly as before.
- `build/admin.js` is gitignored and regenerated at build time; only `src/js/admin.js` is committed.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/tests/phpunit/test-class-live-update.php
+++ b/tests/phpunit/test-class-live-update.php
@@ -1,0 +1,80 @@
+<?php
+namespace WP_Stream;
+
+/**
+ * Class Test_Live_Update
+ *
+ * @package WP_Stream
+ */
+class Test_Live_Update extends WP_StreamTestCase {
+	/**
+	 * Live update instance under test.
+	 *
+	 * @var Live_Update
+	 */
+	protected $live_update;
+
+	public function setUp(): void {
+		parent::setUp();
+
+		$this->live_update = new Live_Update( $this->plugin );
+	}
+
+	public function tearDown(): void {
+		unset( $_POST['action'], $_POST['nonce'], $_POST['checked'], $_POST['heartbeat'], $_POST['user'] );
+
+		parent::tearDown();
+	}
+
+	public function test_enable_live_update_only_updates_current_user() {
+		$attacker_id = $this->factory->user->create( array( 'role' => 'subscriber' ) );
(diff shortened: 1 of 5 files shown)


tests/phpunit/test-class-live-update.php
```diff
@@ -0,0 +1,80 @@
...
+		$_POST['checked']   = 'checked';
+		$_POST['heartbeat'] = 'true';
+
+		try {
+			$this->_handleAjax( 'stream_enable_live_update' );
+		} catch ( \WPAjaxDieContinueException $e ) {
+			// Expected: wp_send_json_error() terminates the AJAX request.
+		}
+
+		$this->assertSame( 'off', get_user_meta( $user_id, $this->live_update->user_meta_key, true ) );
```
PatelUtkarsh, line 84: **test:** This asserts the meta side-effect but not the response. Consider also pinning the denial response so a future refactor can't silently return success without writing meta:

```php
$response = json_decode( $this->_last_response );
$this->assertFalse( $response->success );
```

Minor.

PatelUtkarsh (approved):
Looks good! Left 1 comment.

tests/phpunit/test-class-live-update.php
bartoszgadomski: Good call. Added `$response = json_decode( $this->_last_response )` and asserted `$response->success` is `false` in the denial test, plus the mirror `true` assertion in the current-user test so a refactor can't return success without writing the expected meta. Pushed in 41fcef8a.

PR #1929, opened 2026-07-16
fix(wc): pass order ID, not `WC_Order` object, to Stream log()

### LINKED ISSUE #1928: WooCommerce order-status connector passes a WC_Order object instead of the order ID to Log::log()
# Bug Report

## Expected Behavior

Changing a WooCommerce order's status from wp-admin logs a Stream record with no PHP warnings.

## Actual Behavior

A PHP warning is logged on every WooCommerce order status change:

```
PHP Warning: Object of class Automattic\WooCommerce\Admin\Overrides\Order could not be converted to int in .../stream/classes/class-log.php on line 126
```

This happens because `Connector_WooCommerce::callback_woocommerce_order_status_changed()` fetches a `WC_Order` object where it should fetch just the order ID, then passes that object into `Log::log()`'s `$object_id` parameter (expects an `int`).

Root cause:

https://github.com/xwp/stream/blob/v4.2.2/connectors/class-connector-woocommerce.php#L498

```php
$order_id        = \WC_Order_Factory::get_order( $order_id );
$order_title     = esc_html__( 'Order number', 'stream' ) . ' ' . esc_html( $order_id );
$order_type_name = esc_html__( 'order', 'stream' );

$this->log(
    $message,
    array( /* ... */ ),
    $order_id,   // <-- a WC_Order object, not an int
    'shop_order',
    'updated'
);
```

`WC_Order_Factory::get_order()` returns a `WC_Order` object. The two sibling methods in the same file correctly call `get_order_id()` instead:

- https://github.com/xwp/stream/blob/v4.2.2/connectors/class-connector-woocommerce.php#L399
- https://github.com/xwp/stream/blob/v4.2.2/connectors/class-connector-woocommerce.php#L440

1. Install WooCommerce + Stream, activate both.
2. In wp-admin, change any order's status (e.g. Pending → Completed).
3. Check the PHP error log — the warning above appears.

## Screenshots

N/A — see PHP warning text and stack trace above.

## System Information

- Stream plugin version: 4.2.2
- WordPress version: 7.0.1
- PHP version: 8.4.21
- Browser: N/A (server-side)
- WooCommerce version: 10.9.4

codepuncher:
Fixes #1928.

`Connector_Woocommerce::callback_woocommerce_order_status_changed()` fetched a `WC_Order` object via `WC_Order_Factory::get_order()`, then passed it where an order ID (int) was expected — into `Log::log()`'s `$object_id` parameter. This triggers a PHP warning on every order status change: `Object of class Automattic\WooCommerce\Admin\Overrides\Order could not be converted to int`.

The fix swaps `get_order()` for `get_order_id()`, matching `callback_transition_post_status()` and `callback_deleted_post()` in the same file, which already fetch the ID correctly. No other behaviour changes.

### DIFF
--- a/connectors/class-connector-woocommerce.php
+++ b/connectors/class-connector-woocommerce.php
@@ -495,7 +495,7 @@ class Connector_Woocommerce extends Connector {
 			'stream'
 		);
 
-		$order_id        = \WC_Order_Factory::get_order( $order_id );
+		$order_id        = \WC_Order_Factory::get_order_id( $order_id );
 		$order_title     = esc_html__( 'Order number', 'stream' ) . ' ' . esc_html( $order_id );
 		$order_type_name = esc_html__( 'order', 'stream' );
 


PatelUtkarsh (approved):
Looks good!

Issue #1824, opened 2025-11-28
Exclude by IP address only does not work


slaFFik:
# Bug Report

If I configure the Exclude rule with an IP address, and everything else remains as "any" - Stream will continue logging events happing from that IP address.

## Expected Behavior

If the IP address is added to exclusion list with any other rules, it should be honored.

## Actual Behavior

Stream logs events from the IP address that was added to the ignore list.

## Steps to Reproduce the Problem

1. Set Role, Action, Context to Any.
2. Set the IP address to 127.0.0.1
3. Update your profile in wp-admin/profile.php
4. Observe the new record in Stream table of events

## Screenshots

<img width="862" height="220" alt="Image" src="https://github.com/user-attachments/assets/2ef2bf60-b39a-4dbc-afa3-22d1f4a537e3" />

## System Information

- Stream plugin version: 4.1.1
- WordPress version: 6.8.3
- PHP version: 8.2
- Browser: Chrome
- Computer operating system: MacOS

PR #1932, opened 2026-07-19
fix(log): trim whitespace in comma-separated IP exclude rules

faisalahammad:
Relates to #1824 (does not resolve it — see Summary and review discussion below: the report's actual repro has no comma/whitespace in the IP value, so a different root cause — likely IPv6 `::1` vs `127.0.0.1`, or `REMOTE_ADDR` parsing — remains open in that issue).

## Summary

Exclude rules with a comma-separated multi-IP address value (and all other fields set to Any) did not honor whitespace around the commas. `Log::record_matches_rules()` only split the value with `explode( ',', ... )` and used a strict `in_array()` against it, so a stored `"127.0.0.1, 8.8.8.8"` never matched a real `8.8.8.8` request.

**Note:** this does not address issue #1824 — see the review discussion below for why. This PR only fixes the comma/whitespace handling described here.

### `classes/class-log.php`

**Before:**
```php
if ( 'ip_address' === $exclude_key ) {
	$ip_addresses = explode( ',', $exclude_value );

	if ( in_array( $record['ip_address'], $ip_addresses, true ) ) {
		++$matches_found;
	}
}
```

**After:**
```php
if ( 'ip_address' === $exclude_key ) {
	// Admin form joins multi-IP values with a comma; trim each
	// one so "1.1.1.1, 8.8.8.8" still matches a real client IP.
	$ip_addresses = array_filter( array_map( 'trim', explode( ',', (string) $exclude_value ) ) );

	if ( in_array( $record['ip_address'], $ip_addresses, true ) ) {
		++$matches_found;
	}
}
```

**Why:** the admin Exclude form joins multi-IP values with a comma (`src/js/admin-exclude.js`), so the stored value is a single string. Without trimming, `"127.0.0.1, 8.8.8.8"` becomes `['127.0.0.1', ' 8.8.8.8']` after `explode()` and the strict `in_array()` never matches.

## Repro before the fix

```php
$log->record_matches_rules(
	array( 'ip_address' => '8.8.8.8' ),
	array( 'ip_address' => '127.0.0.1, 8.8.8.8' )
);
// false (expected: true)
```

## Testing

**Multi-IP rule with whitespace**
1. In wp-admin, open Stream -> Settings -> Exclude.
2. Add a rule: leave Author or Role, Context, and Action as Any. Set IP Address to `127.0.0.1, 8.8.8.8` (with or without a space after the comma). Save.
3. Make a request from `8.8.8.8`.

Result: the request is excluded (with the fix). Without the fix, it is still logged.

**Automated tests run:**
- `composer test-one -- --filter='Test_Log::test_ip_address_rule_trims_whitespace_around_commas'` (single + multisite, green)
- `composer test-one -- --filter='Test_Log::test_ip_address_rule_does_not_match_empty_record_ip'` (single + multisite, green)
- `composer test-one -- --filter='Test_Log::test_ip_address_rule_drops_empty_comma_tokens'` (single + multisite, green)
- `composer test-one -- --filter='Test_Log::test_ip_only_exclude_rule_excludes_record'` (single + multisite, green)
- `composer lint-php` and `composer lint-tests` (green)
- CodeRabbit review: 0 findings

# Checklist

## Release Changelog

- Fix: Exclude rules with a comma-separated multi-IP address value now match correctly when the stored value has whitespace around the commas.

## Release Checklist

_N/A — this PR targets `develop`, not `master`; these apply once the fix ships in a release branch._

### DIFF
--- a/classes/class-log.php
+++ b/classes/class-log.php
@@ -242,9 +242,23 @@ class Log {
 			}
 
 			if ( 'ip_address' === $exclude_key ) {
-				$ip_addresses = explode( ',', $exclude_value );
+				// Stored value shape varies: the admin form posts one
+				// comma-joined string per row, while a direct API or test
+				// caller can hand us an array of IPs. Normalize first, then
+				// trim and drop empties so a stored "1.1.1.1, 8.8.8.8" or
+				// ["1.1.1.1", " 8.8.8.8"] both match a real client IP.
+				$ip_addresses = is_array( $exclude_value )
+					? $exclude_value
+					: explode( ',', (string) $exclude_value );
 
-				if ( in_array( $record['ip_address'], $ip_addresses, true ) ) {
+				$ip_addresses = array_filter(
+					array_map( 'trim', $ip_addresses ),
+					function ( $value ) {
+						return '' !== $value;
+					}
+				);
+
+				if ( ! empty( $record['ip_address'] ) && in_array( $record['ip_address'], $ip_addresses, true ) ) {
 					++$matches_found;
 				}
 			} elseif ( $record[ $exclude_key ] === $exclude_value ) {
--- a/classes/class-log.php
+++ b/classes/class-log.php
@@ -242,18 +242,11 @@ class Log {
 			}
 
 			if ( 'ip_address' === $exclude_key ) {
-				// Stored value shape varies: the admin form posts one
-				// comma-joined string per row, while a direct API or test
-				// caller can hand us an array of IPs. Normalize first, then
-				// trim and drop empties so a stored "1.1.1.1, 8.8.8.8" or
-				// ["1.1.1.1", " 8.8.8.8"] both match a real client IP.
-				$ip_addresses = is_array( $exclude_value )
-					? $exclude_value
-					: explode( ',', (string) $exclude_value );
+				// Admin form joins multi-IP values with a comma; trim each
+				// one so "1.1.1.1, 8.8.8.8" still matches a real client IP.
+				$ip_addresses = array_filter( array_map( 'trim', explode( ',', (string) $exclude_value ) ) );
 
-				$ip_addresses = array_filter( array_map( 'trim', $ip_addresses ) );
-
-				if ( ! empty( $record['ip_address'] ) && in_array( $record['ip_address'], $ip_addresses, true ) ) {
+				if ( in_array( $record['ip_address'], $ip_addresses, true ) ) {
 					++$matches_found;
 				}
 			} elseif ( $record[ $exclude_key ] === $exclude_value ) {
--- a/classes/class-log.php
+++ b/classes/class-log.php
@@ -251,12 +251,7 @@ class Log {
 					? $exclude_value
 					: explode( ',', (string) $exclude_value );
 
-				$ip_addresses = array_filter(
-					array_map( 'trim', $ip_addresses ),
-					function ( $value ) {
-						return '' !== $value;
-					}
-				);
+				$ip_addresses = array_filter( array_map( 'trim', $ip_addresses ) );
 
 				if ( ! empty( $record['ip_address'] ) && in_array( $record['ip_address'], $ip_addresses, true ) ) {
 					++$matches_found;
--- a/tests/phpunit/test-class-log.php
+++ b/tests/phpunit/test-class-log.php
@@ -181,4 +181,136 @@ class Test_Log extends WP_StreamTestCase {
 			)
 		);
 	}
+
+	public function test_ip_address_rule_matches_with_whitespace() {
+		// Whitespace around commas (admin form join + user paste) must not
+		// silently fail the IP match. See issue #1824.
+		$this->assertTrue(
+			$this->plugin->log->record_matches_rules(
+				array(
+					'ip_address' => '8.8.8.8',
+				),
+				array(
+					'ip_address' => '1.1.1.1, 8.8.8.8',
+				),
+				'Trailing space after comma does not break the match'
+			)
+		);
+
+		$this->assertTrue(
+			$this->plugin->log->record_matches_rules(
+				array(
+					'ip_address' => '8.8.8.8',
+				),
+				array(
+					'ip_address' => '8.8.8.8 ',
+				),
+				'Trailing space on a single-IP rule still matches'
+			)
+		);
+
+		$this->assertFalse(
+			$this->plugin->log->record_matches_rules(
+				array(
+					'ip_address' => '',
+				),
+				array(
+					'ip_address' => '1.1.1.1',
+				),
+				'Empty record IP never matches an IP-only rule'
+			)
+		);
+
+		// Empty tokens between commas (e.g. user-pasted "1.1.1.1, ,") must be
+		// dropped, not treated as a valid match against the empty record IP.
+		$this->assertTrue(
+			$this->plugin->log->record_matches_rules(
+				array(
+					'ip_address' => '1.1.1.1',
+				),
+				array(
+					'ip_address' => '1.1.1.1, ,',
+				),
+				'Empty comma-separated tokens are dropped before matching'
+			)
+		);
+
+		// Stored value may already be an array (direct API callers, tests).
+		// Both shapes must match.
+		$this->assertTrue(
+			$this->plugin->log->record_matches_rules(
+				array(
+					'ip_address' => '8.8.8.8',
+				),
+				array(
+					'ip_address' => array( '127.0.0.1', '8.8.8.8' ),
+				),
+				'Array-shaped IP rule matches the second entry'
+			)
+		);
+
+		$this->assertTrue(
+			$this->plugin->log->record_matches_rules(
+				array(
+					'ip_address' => '8.8.8.8',
+				),
+				array(
+					'ip_address' => array( ' 8.8.8.8 ' ),
+				),
+				'Array-shaped IP rule trims whitespace per entry'
+			)
+		);
+	}
+
+	public function test_ip_only_exclude_rule_excludes_record() {
+		// End-to-end coverage for the bug in #1824. Shape mirrors the
+		// parallel-array rule format produced by both the wp-admin Exclude
+		// list and the stream/create-exclusion-rule ability.
+		$this->plugin->settings->options['exclude_rules'] = array(
+			'exclude_row'    => array( 0 => '' ),
+			'author_or_role' => array( 0 => '' ),
+			'connector'      => array( 0 => '' ),
+			'context'        => array( 0 => '' ),
+			'action'         => array( 0 => '' ),
+			'ip_address'     => array( 0 => '127.0.0.1' ),
+		);
+
+		$user = $this->factory->user->create_and_get();
+		$user->add_role( 'administrator' );
+
+		$this->assertTrue(
+			$this->plugin->log->is_record_excluded(
+				'users',
+				'profile',
+				'updated',
+				$user,
+				'127.0.0.1'
+			),
+			'IP-only rule excludes a record from the matching IP'
+		);
+
+		$this->assertFalse(
+			$this->plugin->log->is_record_excluded(
+				'users',
+				'profile',
+				'updated',
+				$user,
+				'8.8.8.8'
+			),
+			'IP-only rule does not exclude a record from a different IP'
+		);
+
+		// Whitespace in the stored IP value must still match.
+		$this->plugin->settings->options['exclude_rules']['ip_address'][0] = '127.0.0.1, 8.8.8.8';
+		$this->assertTrue(
+			$this->plugin->log->is_record_excluded(
+				'users',
+				'profile',
+				'updated',
+				$user,
+				'8.8.8.8'
+			),
+			'Comma-joined IP list with whitespace matches the second entry'
+		);
+	}
 }
--- a/tests/phpunit/test-class-log.php
+++ b/tests/phpunit/test-class-log.php
@@ -182,9 +182,9 @@ class Test_Log extends WP_StreamTestCase {
 		);
 	}
 
-	public function test_ip_address_rule_matches_with_whitespace() {
+	public function test_ip_address_rule_trims_whitespace_around_commas() {
 		// Whitespace around commas (admin form join + user paste) must not
-		// silently fail the IP match. See issue #1824.
+		// silently fail the IP match.
 		$this->assertTrue(
 			$this->plugin->log->record_matches_rules(
 				array(
@@ -192,9 +192,9 @@ class Test_Log extends WP_StreamTestCase {
 				),
 				array(
 					'ip_address' => '1.1.1.1, 8.8.8.8',
-				),
-				'Trailing space after comma does not break the match'
-			)
+				)
+			),
+			'Trailing space after comma does not break the match'
 		);
 
 		$this->assertTrue(
@@ -204,11 +204,13 @@ class Test_Log extends WP_StreamTestCase {
 				),
 				array(
 					'ip_address' => '8.8.8.8 ',
-				),
-				'Trailing space on a single-IP rule still matches'
-			)
+				)
+			),
+			'Trailing space on a single-IP rule still matches'
 		);
+	}
 
+	public function test_ip_address_rule_does_not_match_empty_record_ip() {
 		$this->assertFalse(
 			$this->plugin->log->record_matches_rules(
 				array(
@@ -216,11 +218,13 @@ class Test_Log extends WP_StreamTestCase {
 				),
 				array(
 					'ip_address' => '1.1.1.1',
-				),
-				'Empty record IP never matches an IP-only rule'
-			)
+				)
+			),
+			'Empty record IP never matches an IP-only rule'
 		);
+	}
 
+	public function test_ip_address_rule_drops_empty_comma_tokens() {
 		// Empty tokens between commas (e.g. user-pasted "1.1.1.1, ,") must be
 		// dropped, not treated as a valid match against the empty record IP.
 		$this->assertTrue(
@@ -230,42 +234,17 @@ class Test_Log extends WP_StreamTestCase {
 				),
 				array(
 					'ip_address' => '1.1.1.1, ,',
-				),
-				'Empty comma-separated tokens are dropped before matching'
-			)
-		);
-
-		// Stored value may already be an array (direct API callers, tests).
-		// Both shapes must match.
-		$this->assertTrue(
-			$this->plugin->log->record_matches_rules(
-				array(
-					'ip_address' => '8.8.8.8',
-				),
-				array(
-					'ip_address' => array( '127.0.0.1', '8.8.8.8' ),
-				),
-				'Array-shaped IP rule matches the second entry'
-			)
-		);
-
-		$this->assertTrue(
-			$this->plugin->log->record_matches_rules(
-				array(
-					'ip_address' => '8.8.8.8',
-				),
-				array(
-					'ip_address' => array( ' 8.8.8.8 ' ),
-				),
-				'Array-shaped IP rule trims whitespace per entry'
-			)
+				)
+			),
+			'Empty comma-separated tokens are dropped before matching'
 		);
 	}
 
 	public function test_ip_only_exclude_rule_excludes_record() {
-		// End-to-end coverage for the bug in #1824. Shape mirrors the
-		// parallel-array rule format produced by both the wp-admin Exclude
-		// list and the stream/create-exclusion-rule ability.
+		// End-to-end coverage for the comma/whitespace exclude-rule fix.
+		// Shape mirrors the parallel-array rule format produced by both the
+		// wp-admin Exclude list and the stream/create-exclusion-rule ability.
+		// Does not cover the separate IPv6/proxy-chain cause of issue #1824.
 		$this->plugin->settings->options['exclude_rules'] = array(
 			'exclude_row'    => array( 0 => '' ),
 			'author_or_role' => array( 0 => '' ),


classes/class-log.php
```diff
@@ -242,9 +242,23 @@ public function record_matches_rules( $record, $exclude_rules ) {
...
+				$ip_addresses = is_array( $exclude_value )
+					? $exclude_value
+					: explode( ',', (string) $exclude_value );
+
+				$ip_addresses = array_filter(
+					array_map( 'trim', $ip_addresses ),
+					function ( $value ) {
+						return '' !== $value;
+					}
+				);
```
shadyvb: nitpick: We can drop the anonymous function here and use the native filtering logic of `array_filter()`.

```suggestion
				$ip_addresses = array_filter( array_map( 'trim', $ip_addresses ) );
```

shadyvb (commented):
Looks good! With one nitpick about the unnecessary anonymous function usage with `array_filter()`.

classes/class-log.php
faisalahammad: Fixed. Dropped the anonymous callback and used native array_filter, which removes empty entries after trim. Both Test_Log tests still pass (single + multisite).

tests/phpunit/test-class-log.php
```diff
@@ -181,4 +181,136 @@ public function test_can_match_record_id_address() {
...
+			$this->plugin->log->is_record_excluded(
+				'users',
+				'profile',
+				'updated',
+				$user,
+				'8.8.8.8'
+			),
+			'Comma-joined IP list with whitespace matches the second entry'
+		);
+	}
```
PatelUtkarsh: **Blocker:** this test does not show the defect in #1824.

I changed `classes/class-log.php` back to the `origin/develop` version. Then I ran this test with no changes. The test failed only at line 313, which is the whitespace assertion at the end.

The first two assertions use the same conditions as the issue report: one rule that has only an IP address (`127.0.0.1`), and all the other fields set to Any. Both of these assertions passed on `develop` without the correction.

Thus there are two possible causes. The PR must show which cause is correct.

1. Whitespace is not the cause of #1824. This PR corrects a different defect, and the defect in the report stays open.
2. The data in the test setup is not the same as the stored data.

I think that cause 2 is more probable. This test writes directly to `$this->plugin->settings->options['exclude_rules']` in memory. But the operational code in `Log::is_record_excluded()` (`classes/class-log.php:185-190`) reads `get_site_option( 'wp_stream_network' )` when the plugin is network-activated. Also, in the usual wp-admin flow, `Settings::sanitize_settings()` processes the value before storage. This test uses neither path, thus it does not include the storage step.

The person who made the report uses version 4.1.1 on a single site. Thus the multisite path is probably not the cause. But the save path is still not tested.

Please add a test that saves the rule through the settings save path (or with `update_option`), and then makes the assertion about the exclusion. Then make sure that this new test fails on `develop`. If the new test passes on `develop`, then #1824 has a different cause. In that condition, remove `Fixes #1824` from this PR.

classes/class-log.php
```diff
@@ -242,9 +242,18 @@ public function record_matches_rules( $record, $exclude_rules ) {
...
-
-				if ( in_array( $record['ip_address'], $ip_addresses, true ) ) {
+				// Stored value shape varies: the admin form posts one
+				// comma-joined string per row, while a direct API or test
+				// caller can hand us an array of IPs. Normalize first, then
+				// trim and drop empties so a stored "1.1.1.1, 8.8.8.8" or
+				// ["1.1.1.1", " 8.8.8.8"] both match a real client IP.
+				$ip_addresses = is_array( $exclude_value )
+					? $exclude_value
+					: explode( ',', (string) $exclude_value );
```
PatelUtkarsh: **Warning:** I could not find operational code that gives an array to this parameter. Thus this branch is not necessary.

I examined each path that supplies `$exclude_value`:

- `abilities/class-ability-create-exclusion-rule.php:138` makes a `(string)` cast. Line 147 does a check with `FILTER_VALIDATE_IP`. Thus the ability always stores a scalar value.
- `Settings::render_field()` (`classes/class-settings.php:1100`) reads a scalar value.
- `src/js/admin-exclude.js:378-383` joins the selected values into one string with commas.
- `Log::is_record_excluded():197` gives the stored value to the function with no change.

Only the test in this PR gives an array to this function. Thus the test is the only proof that the branch is necessary, but the test is also a part of this PR.

Do one of these two things. Show operational code that gives an array. Or remove this branch and the two array assertions in the test (lines 240-262). The `(string)` cast is sufficient for the PHP 8 `TypeError` that the description mentions.

classes/class-log.php
```diff
@@ -242,9 +242,18 @@ public function record_matches_rules( $record, $exclude_rules ) {
...
 
 			if ( 'ip_address' === $exclude_key ) {
-				$ip_addresses = explode( ',', $exclude_value );
-
-				if ( in_array( $record['ip_address'], $ip_addresses, true ) ) {
+				// Stored value shape varies: the admin form posts one
+				// comma-joined string per row, while a direct API or test
+				// caller can hand us an array of IPs. Normalize first, then
+				// trim and drop empties so a stored "1.1.1.1, 8.8.8.8" or
+				// ["1.1.1.1", " 8.8.8.8"] both match a real client IP.
```
PatelUtkarsh: **Nit:** the comment has five lines, but the code has four lines.

The comment also tells about a "direct API or test caller". I could not find such a caller. See my other comment about the array branch.

Two lines are sufficient: the admin form joins the values with commas, thus the code must remove the spaces before the strict comparison.

tests/phpunit/test-class-log.php
```diff
@@ -181,4 +181,136 @@ public function test_can_match_record_id_address() {
 			)
 		);
 	}
+
+	public function test_ip_address_rule_matches_with_whitespace() {
```
PatelUtkarsh: **Nit:** this test method makes assertions about four different behaviors, but the name tells about only one behavior. The four behaviors are: removal of spaces, an empty record IP, removal of empty items, and an array value.

If the test fails, the name of the test will point to the incorrect behavior. Divide this test into more than one test. Or give it a more general name, for example `test_ip_address_rule_normalization`.

tests/phpunit/test-class-log.php
PatelUtkarsh: **Blocker:** this test does not show the defect in #1824.

I changed `classes/class-log.php` back to the `origin/develop` version. Then I ran this test with no changes. The test failed only at line 313, which is the whitespace assertion at the end.

The first two assertions use the same conditions as the issue report: one rule that has only an IP address (`127.0.0.1`), and all the other fields set to Any. Both of these assertions passed on `develop` **without** the correction in this PR. Thus this test does not show the defect in the report.

**What I think the cause is**

This test gives a clean address directly to `is_record_excluded()`. But in operation the address comes from `Plugin::$client_ip_address`, and `classes/class-plugin.php:221` makes that value:

```php
$this->client_ip_address = wp_stream_filter_input( INPUT_SERVER, 'REMOTE_ADDR', FILTER_VALIDATE_IP );
```

Some hosts and some plugins copy the `X-Forwarded-For` value into `REMOTE_ADDR` with no change. Then `REMOTE_ADDR` holds more than one address, separated by commas. I wrote a temporary probe test to examine the result:

```
'203.0.113.9'               => '203.0.113.9'
'203.0.113.9, 70.41.3.18'   => false      <-- chain with a space
'203.0.113.9,70.41.3.18'    => false      <-- chain with no space
'203.0.113.9:51234'         => false      <-- address with a port
'::1'                       => '::1'
```

A chain gives `false`, not the first address. `Filter_Input::filter()` at `classes/class-filter-input.php:114-117` causes this: when a validator fails, the code sets the value to `false`.

Two results follow. `Log::log():136` writes `'ip' => (string) $ip_address`, thus the IP column is empty for every record. And `Log::is_record_excluded():182` puts the same `false` into `$record['ip_address']`, thus a rule that has only an IP address can never agree with the record.

I made a test of the full path:

```
rule = '203.0.113.9', record ip = false                      => not excluded
rule = '203.0.113.9', record ip = '203.0.113.9, 70.41.3.18'  => not excluded
rule = '203.0.113.9', record ip = '203.0.113.9'              => excluded  (control)
```

This agrees with the report: the rule does nothing, and no error message occurs.

**How to confirm**

I cannot confirm that the environment of the reporter has this condition. Please ask the reporter to open Stream, then Records, and look at the IP column. If the column is empty for all records, this defect is the cause. If the column shows an address that is not `127.0.0.1`, then `REMOTE_ADDR` holds the address of the proxy or the container, the rule is correct but does not agree, and the correction is documentation.

**What to do**

1. Remove `Fixes #1824` from this PR. Keep this PR for the whitespace defect only. Make a new issue for the proxy chain defect.
2. Or also correct `classes/class-plugin.php:221`, and add a test that sets `$_SERVER['REMOTE_ADDR']` to a chain and then makes an assertion about the exclusion.

Option 1 is better, because the two defects are in different files and each one needs its own test.

If you correct the proxy defect, correct it at the source in `classes/class-plugin.php:221`: take the first item of the chain, remove the spaces, then do the validation. Do not read the `X-Forwarded-For` header, because a client can set that header. Commit 38683c13 removed that header for this reason.

classes/class-log.php
```diff
@@ -242,9 +242,18 @@ public function record_matches_rules( $record, $exclude_rules ) {
...
+				$ip_addresses = is_array( $exclude_value )
+					? $exclude_value
+					: explode( ',', (string) $exclude_value );
+
+				$ip_addresses = array_filter( array_map( 'trim', $ip_addresses ) );
+
+				if ( ! empty( $record['ip_address'] ) && in_array( $record['ip_address'], $ip_addresses, true ) ) {
```
PatelUtkarsh: **Warning:** the `! empty( $record['ip_address'] )` condition has no effect on the result.

When the code gets to this line, two things are already true. The condition `! isset( $record[ $exclude_key ] )` at line 240 passed. And line 254 removed all the empty items from `$ip_addresses`. Thus an empty record IP address, or a record IP address of `'0'`, cannot agree with an item in the list. The result is the same with the condition and without it.

But an empty record IP address does occur in operation. When `REMOTE_ADDR` holds a proxy chain, `classes/class-plugin.php:221` gives `false`, and that value goes into the record. See my blocker comment on the test file.

Thus, if you added this condition for that reason, the condition is in the wrong place. It hides the problem here, but the correction belongs at the source in `classes/class-plugin.php:221`.

If the condition is only defensive, tell that in a comment. At this time, `test_ip_address_rule_matches_with_whitespace` (lines 213-221 of the test file) makes an assertion about this condition. Thus a subsequent reader will think that the condition is necessary.

classes/class-log.php
PatelUtkarsh: **Warning:** I could not find operational code that gives an array to this parameter. Thus this branch is not necessary.

I examined each path that supplies `$exclude_value`:

- `abilities/class-ability-create-exclusion-rule.php:138` makes a `(string)` cast. Line 147 does a check with `FILTER_VALIDATE_IP`. Thus the ability always stores a scalar value.
- `Settings::render_field()` (`classes/class-settings.php:1100`) reads a scalar value.
- `src/js/admin-exclude.js:378-383` joins the selected values into one string with commas.
- `Log::is_record_excluded():197` gives the stored value to the function with no change.

Only the test in this PR gives an array to this function. Thus the test is the only proof that the branch is necessary, but the test is also a part of this PR.

Do one of these two things. Show operational code that gives an array. Or remove this branch and the two array assertions in the test (lines 240-262). The `(string)` cast is sufficient for the PHP 8 `TypeError` that the description mentions.

classes/class-log.php
PatelUtkarsh: **Nit:** the comment has five lines, but the code has four lines.

The comment also tells about a "direct API or test caller". I could not find such a caller. See my other comment about the array branch.

Two lines are sufficient: the admin form joins the values with commas, thus the code must remove the spaces before the strict comparison.

classes/class-log.php
PatelUtkarsh: **Warning:** the `! empty( $record['ip_address'] )` condition has no effect on the result.

When the code gets to this line, two things are already true. The condition `! isset( $record[ $exclude_key ] )` at line 240 passed. And line 254 removed all the empty items from `$ip_addresses`. Thus an empty record IP address, or a record IP address of `'0'`, cannot agree with an item in the list. The result is the same with the condition and without it.

But an empty record IP address does occur in operation. When `REMOTE_ADDR` holds a proxy chain, `classes/class-plugin.php:221` gives `false`, and that value goes into the record. See my blocker comment on the test file.

Thus, if you added this condition for that reason, the condition is in the wrong place. It hides the problem here, but the correction belongs at the source in `classes/class-plugin.php:221`.

If the condition is only defensive, tell that in a comment. At this time, `test_ip_address_rule_matches_with_whitespace` (lines 213-221 of the test file) makes an assertion about this condition. Thus a subsequent reader will think that the condition is necessary.

classes/class-log.php
PatelUtkarsh: **Warning:** I could not find operational code that gives an array to this parameter. Thus this branch is not necessary.

I examined each path that supplies `$exclude_value`:

- `abilities/class-ability-create-exclusion-rule.php:138` makes a `(string)` cast. Line 147 does a check with `FILTER_VALIDATE_IP`. Thus the ability always stores a scalar value.
- `Settings::render_field()` (`classes/class-settings.php:1100`) reads a scalar value.
- `src/js/admin-exclude.js:378-383` joins the selected values into one string with commas.
- `Log::is_record_excluded():197` gives the stored value to the function with no change.

Only the test in this PR gives an array to this function. Thus the test is the only proof that the branch is necessary, but the test is also a part of this PR.

Do one of these two things. Show operational code that gives an array. Or remove this branch and the two array assertions in the test (lines 240-262). The `(string)` cast is sufficient for the PHP 8 `TypeError` that the description mentions.

tests/phpunit/test-class-log.php
PatelUtkarsh, line 294: **Blocker:** this test does not show the defect in #1824, and I do not think that whitespace is the cause of that report.

**Evidence 1: the screenshot in the issue**

The screenshot shows one rule row. Author or Role, Context, and Action all show the placeholder text. The IP Address cell has exactly one select2 tag: `× 127.0.0.1`.

One tag. No comma. Thus the stored value is a bare `127.0.0.1`, and it has no whitespace to remove and no second address. The correction in this PR operates only on a value that has a comma or a space. Thus it cannot change the result for this reporter.

**Evidence 2: the test passes without the correction**

I changed `classes/class-log.php` back to the `origin/develop` version. Then I ran this test with no changes. The test failed only at line 313, which is the whitespace assertion at the end. The first two assertions, which use the same conditions as the report, both passed on `develop` **without** the correction in this PR.

**Evidence 3: I reproduced a different cause on macOS**

The reporter uses macOS. On macOS, the name `localhost` gives the IPv6 address first:

```
$ dscacheutil -q host -a name localhost
  ipv6_address: ::1        <-- first
  ip_address: 127.0.0.1
```

Thus, when the web server accepts IPv6 and the user opens `http://localhost`, the value of `REMOTE_ADDR` is `::1`. I made a test with a small PHP server that shows `REMOTE_ADDR`:

```
Server on localhost (accepts IPv6):
  http://localhost:8899/   => REMOTE_ADDR=::1

Server on 0.0.0.0 (IPv4 only):
  http://localhost:8899/   => REMOTE_ADDR=127.0.0.1
  http://127.0.0.1:8899/   => REMOTE_ADDR=127.0.0.1
```

The value depends on the address that the server accepts, not on the text that the user writes. Thus a user can open `localhost`, see a correct rule of `127.0.0.1`, and get `::1` in the record.

A probe test with a rule of `127.0.0.1` gives these results:

```
REMOTE_ADDR=127.0.0.1          excluded=true
REMOTE_ADDR=::1                excluded=false
REMOTE_ADDR=::ffff:127.0.0.1   excluded=false
REMOTE_ADDR=172.18.0.1         excluded=false
```

This agrees with the report: the rule does nothing, and no error message occurs.

Note that this is not a defect in the comparison. `::1` and `127.0.0.1` are different addresses, thus a strict comparison is correct. The defect is that Stream gives the user no feedback. The user writes an address that looks correct, Stream records a different address, but the UI does not show this difference.

**How to reproduce this without the reporter**

1. Start WordPress on a server that accepts IPv6. The default Apache on macOS does this.
2. Open the site with `http://localhost`, not with `http://127.0.0.1`.
3. Add an exclude rule. Set the IP address to `127.0.0.1`. Set all the other fields to Any. Save.
4. Open `wp-admin/profile.php`. Select Update Profile.
5. Open Stream, then Records. The record is present. The IP column shows `::1`.
6. Change the rule to `::1`. Do step 4 again. The record is absent.

Step 6 is the proof. If the rule operates with `::1` but not with `127.0.0.1`, this is the cause.

**A second, different defect that I found**

This one is real, but it is probably not the defect of this reporter. `classes/class-plugin.php:221` resolves the client IP address with `FILTER_VALIDATE_IP`. Some hosts copy the `X-Forwarded-For` value into `REMOTE_ADDR` with no change, thus `REMOTE_ADDR` holds more than one address:

```
'203.0.113.9'               => '203.0.113.9'
'203.0.113.9, 70.41.3.18'   => false
'203.0.113.9:51234'         => false
```

A chain gives `false`, not the first address, because of `Filter_Input::filter()` at `classes/class-filter-input.php:114-117`. Then `Log::log():136` writes an empty IP column for every record, and `Log::is_record_excluded():182` puts `false` into the record. Thus no IP rule can ever agree. This gives the same symptom, but only on a site that has a proxy.

**What to do**

1. Remove `Fixes #1824` from this PR. Keep this PR for the whitespace defect, which is real but is a different defect.
2. Do the six steps above. If step 6 gives the result that I describe, make a new issue for the true cause.

The whitespace correction is good. My objection is only to the statement that it closes #1824.

classes/class-log.php
PatelUtkarsh: **Warning:** the `! empty( $record['ip_address'] )` condition has no effect on the result.

When the code gets to this line, two things are already true. The condition `! isset( $record[ $exclude_key ] )` at line 240 passed. And line 254 removed all the empty items from `$ip_addresses`. Thus an empty record IP address, or a record IP address of `'0'`, cannot agree with an item in the list. The result is the same with the condition and without it.

But an empty record IP address does occur in operation. When `REMOTE_ADDR` holds a proxy chain, `classes/class-plugin.php:221` gives `false`, and that value goes into the record. See my blocker comment on the test file.

Thus, if you added this condition for that reason, the condition is in the wrong place. It hides the problem here, but the correction belongs at the source in `classes/class-plugin.php:221`.

If the condition is only defensive, tell that in a comment. At this time, `test_ip_address_rule_matches_with_whitespace` (lines 213-221 of the test file) makes an assertion about this condition. Thus a subsequent reader will think that the condition is necessary.

classes/class-log.php
PatelUtkarsh: **Warning:** I could not find operational code that gives an array to this parameter. Thus this branch is not necessary.

I examined each path that supplies `$exclude_value`:

- `abilities/class-ability-create-exclusion-rule.php:138` makes a `(string)` cast. Line 147 does a check with `FILTER_VALIDATE_IP`. Thus the ability always stores a scalar value.
- `Settings::render_field()` (`classes/class-settings.php:1100`) reads a scalar value.
- `src/js/admin-exclude.js:378-383` joins the selected values into one string with commas.
- `Log::is_record_excluded():197` gives the stored value to the function with no change.

Only the test in this PR gives an array to this function. Thus the test is the only proof that the branch is necessary, but the test is also a part of this PR.

Do one of these two things. Show operational code that gives an array. Or remove this branch and the two array assertions in the test (lines 240-262). The `(string)` cast is sufficient for the PHP 8 `TypeError` that the description mentions.

classes/class-log.php
PatelUtkarsh: **Nit:** the comment has five lines, but the code has four lines.

The comment also tells about a "direct API or test caller". I could not find such a caller. See my other comment about the array branch.

Two lines are sufficient: the admin form joins the values with commas, thus the code must remove the spaces before the strict comparison.

tests/phpunit/test-class-log.php
PatelUtkarsh: **Nit:** this test method makes assertions about four different behaviors, but the name tells about only one behavior. The four behaviors are: removal of spaces, an empty record IP, removal of empty items, and an array value.

If the test fails, the name of the test will point to the incorrect behavior. Divide this test into more than one test. Or give it a more general name, for example `test_ip_address_rule_normalization`.

PatelUtkarsh (changes requested):
@faisalahammad Thanks you for your contribution! 

The screenshot in https://github.com/xwp/stream/issues/1824 shows a rule value that has no whitespace and no comma, thus the mechanism in this PR cannot apply to that report. I also reproduced a different cause on macOS that gives the same symptom. See the blocker comment on the test file.

Note: PR comments are created with help of LLM and manually reviewed before posting it.

tests/phpunit/test-class-log.php
faisalahammad: Checked both points in your evidence. IPv6 loopback (::1) is not normalized to 127.0.0.1 anywhere in the code, so a rule for 127.0.0.1 will not match a record stored as ::1, with or without this fix. And Filter_Input::filter() in class-filter-input.php returns false on a comma-chain or port-suffixed REMOTE_ADDR, with no fallback. Neither is touched by this PR.

Removed "Fixes #1824" from the PR title and body. This PR now only claims the whitespace/comma fix. I will leave the new issue for the IPv6/proxy-chain cause to you or the reporter, since you already have the repro steps written out here.

classes/class-log.php
faisalahammad: Removed. The array_filter(array_map('trim', ...)) two lines above already drops every falsy value from the list before in_array() runs, so the !empty() check never changed the result for any input. Pushed in 3689a05d.

classes/class-log.php
faisalahammad: Removed the is_array() branch and the two array-shaped test assertions. Checked every writer of exclude_rules (wp-admin form, and the create-exclusion-rule ability) and every caller of record_matches_rules()/is_record_excluded() again, none passes an array. Kept the (string) cast, since you noted that alone covers the PHP 8 case. Pushed in 3689a05d.

classes/class-log.php
faisalahammad: Shortened to 2 lines and dropped the "direct API or test caller" line, you are right that no such caller exists. Pushed in 3689a05d.

tests/phpunit/test-class-log.php
faisalahammad: Split into 3 tests: test_ip_address_rule_trims_whitespace_around_commas, test_ip_address_rule_does_not_match_empty_record_ip, test_ip_address_rule_drops_empty_comma_tokens. Each name says what it checks. Pushed in 3689a05d.

PatelUtkarsh:
Since this doesn't fix the original issue at https://github.com/xwp/stream/issues/1824, can you share the reproduction steps? @faisalahammad

faisalahammad:
The issue itself has the repro steps (steps 1-4 in the description): Role/Action/Context set to Any, IP set to 127.0.0.1, then an admin request from localhost on macOS.

Like you found, macOS resolves localhost to IPv6 ::1 first. A rule stored as 127.0.0.1 will not match a record logged as ::1, since the match is a strict string compare with no normalization. That is separate from what this PR fixes.

I do not have a repro beyond that. Agree a new issue is the right place for the IPv6 loopback case and the Filter_Input::filter() proxy issue you found. I can file it, or you can, whichever is easier for you.

PR #1963, opened 2026-08-11
Fix fatal in role_can_view() when cap check fires before init

### LINKED ISSUE #1962: Fatal TypeError in `role_can_view()` when a capability check fires before `init` priority 9 (PHP 8)
# Bug Report

## Expected Behavior
I expect no fatal errors.

Describe what should happen.

## Actual Behavior

`Admin::role_can_view()` fatals with an uncaught `TypeError` on PHP 8 when any code checks the `view_stream` capability for a user with roles **before** Stream constructs its `Settings` object. The capability filters are registered earlier than the object they depend on, leaving a window on every request where a cap check crashes the site.

```
PHP Fatal error: Uncaught TypeError: in_array(): Argument #2 ($haystack) must be of type array, null given
  in wp-content/plugins/stream/classes/class-admin.php:1756
#0 class-admin.php(1756): in_array('subscriber', NULL, true)
#1 class-admin.php(...): WP_Stream\Admin->role_can_view('subscriber')
#2 class-wp-hook.php(341): WP_Stream\Admin->filter_user_caps(Array, Array, Array, Object(WP_User))
...
#10 wp-settings.php(593): do_action('plugins_loaded')
```

(Line 1756 in 4.3.0;)

## Root cause

Ordering problem between hook registration and dependency construction:

1. `Admin::__construct()` registers `user_has_cap` / `role_has_cap` filters immediately, while the plugin file is being loaded (`class-admin.php`, `filter_user_caps` / `filter_role_caps`).
2. `Plugin::init()`, which constructs the `Settings` object, only runs on `init` priority 9 (`class-plugin.php`).
3. Between plugin load and `init` 9, `$this->plugin->settings` is `null`, so in `role_can_view()`:

```php
if ( in_array( $role, $this->plugin->settings->options['general_role_access'], true ) ) {
```

the null-safe-less chain evaluates to `null`, and `in_array()` with a `null` haystack is a fatal `TypeError` on PHP 8+.

The crash requires the checked user to have at least one role (the role loop in `filter_user_caps()` is what reaches `role_can_view()`), so it fires on logged-in requests and not for anonymous visitors — i.e. it takes down `wp-admin` and `admin-ajax.php` for every logged-in user, including administrators.

## Real-world trigger

This is not theoretical. Patchstack's firewall (and by extension any security/firewall/MU plugin doing early capability checks) evaluates rules on `plugins_loaded` (priority 10). One of their rule types calls `current_user_can( 'view_stream' )` — a vPatch rule shipped for the Stream authorization issue fixed in 4.2.1/4.2.2 — which lands exactly inside the window and fatals the whole admin. Sites running Stream + Patchstack lost wp-admin access with no code change on their side, since firewall rules are pushed remotely.

Any of these also trigger it:

- an mu-plugin or drop-in checking `current_user_can()`/`user_can()` for a cap that maps to `view_stream` on `plugins_loaded`, `setup_theme`, or early `init` (< 9);
- another plugin calling `user_can( $user, 'view_stream' )` early.

## Steps to reproduce

1. Fresh install, PHP 8.0+, Stream active (reproduced on 4.1.1 and 4.3.0).
2. Add an mu-plugin:

```php
<?php
add_action(
	'plugins_loaded',
	function () {
		user_can( 1, 'view_stream' ); // any existing user that has a role
	},
	20
);
```

3. Load `/wp-admin/`.

**Expected:** page loads; the early check simply returns false.
**Actual:** HTTP 500, uncaught `TypeError` as above.

## Suggested fix

Guard the options access in `role_can_view()` and fail closed:

```php
private function role_can_view( $role ) {
	$allowed_roles = $this->plugin->settings->options['general_role_access'] ?? array();

	return in_array( $role, (array) $allowed_roles, true );
}
```

`??` is null-safe across the whole chain (including `settings` itself being `null`), and denying access during the pre-`init` window is the correct behavior for a cap grant that cannot be evaluated yet.

A more structural alternative: register the `user_has_cap`/`role_has_cap` filters only after `Settings` exists (e.g. from `Plugin::init()` on `init` 9), so the filters never run without their dependency.

## Environment

- Stream: reproduced on 4.1.1 and 4.3.0 (latest); code unchanged on `master` at time of writing
- WordPress: 7.0.2
- PHP: 8.4 (any 8.x fatals; PHP 7 only emitted a warning, which is why this went unnoticed)

andreilupu:
Fixes #1962.

The `user_has_cap`/`role_has_cap` filters are registered in the `Admin` constructor, but the `Settings` object they dereference is only constructed in `Plugin::init()` on `init` priority 9. Any capability check for `view_stream` between `plugins_loaded` and `init` 9 (e.g. a security plugin evaluating firewall rules on `plugins_loaded`) reaches `role_can_view()` with a null options chain, and `in_array()` with a `null` haystack fatals on PHP 8+ — taking down `wp-admin` and `admin-ajax.php` for every logged-in user.

This PR null-coalesces the options access in `role_can_view()` and denies access during the pre-`init` window instead of fataling. Post-`init` behavior is unchanged. Includes regression tests for both paths (early check denied without fatal, allowed role still granted the cap).

# Checklist

## Release Changelog

- Fix: Fatal `TypeError` in `role_can_view()` when the `view_stream` capability is checked before `init` priority 9 (e.g. by security/firewall plugins on `plugins_loaded`) on PHP 8+.

### DIFF
--- a/classes/class-admin.php
+++ b/classes/class-admin.php
@@ -1748,16 +1748,20 @@ class Admin {
 	/**
 	 * Check if a particular role has access
 	 *
+	 * The user_has_cap/role_has_cap filters that call this are registered in the
+	 * constructor, but the Settings object is not constructed until init priority 9.
+	 * A capability check fired before then (e.g. by a security plugin evaluating
+	 * firewall rules on plugins_loaded) must be denied rather than fatal on the
+	 * null options chain.
+	 *
 	 * @param string $role  User role.
 	 *
 	 * @return bool
 	 */
 	private function role_can_view( $role ) {
-		if ( in_array( $role, $this->plugin->settings->options['general_role_access'], true ) ) {
-			return true;
-		}
+		$allowed_roles = $this->plugin->settings->options['general_role_access'] ?? array();
 
-		return false;
+		return in_array( $role, (array) $allowed_roles, true );
 	}
 
 	/**
--- a/tests/phpunit/test-class-admin.php
+++ b/tests/phpunit/test-class-admin.php
@@ -73,6 +73,45 @@ class Test_Admin extends WP_StreamTestCase {
 		$this->assertInstanceOf( '\WP_Stream\Export', $this->admin->export );
 	}
 
+	/**
+	 * The user_has_cap filter is registered in the Admin constructor, but the
+	 * Settings object is only built on init priority 9. A capability check for
+	 * the view cap fired before then (e.g. a firewall plugin on plugins_loaded)
+	 * must be denied, not fatal on the null options chain.
+	 */
+	public function test_filter_user_caps_before_settings_initialized() {
+		$settings               = $this->plugin->settings;
+		$this->plugin->settings = null;
+
+		$user    = get_user_by( 'id', $this->admin_user_id );
+		$allcaps = $this->admin->filter_user_caps(
+			array(),
+			array( $this->admin->view_cap ),
+			array( $this->admin->view_cap, $this->admin_user_id ),
+			$user
+		);
+
+		$this->plugin->settings = $settings;
+
+		$this->assertArrayNotHasKey( $this->admin->view_cap, $allcaps );
+	}
+
+	/**
+	 * Once Settings exists, the view cap is granted to allowed roles as before.
+	 */
+	public function test_filter_user_caps_grants_view_cap_to_allowed_role() {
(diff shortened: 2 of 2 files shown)


shadyvb (approved):
Looks good, thanks for submitting the PR!

PR #1967, opened 2026-08-21
Add AI Client connector (XWPENG-20)

shadyvb:
Fixes XWPENG-20.

Adds a bundled **AI Client** connector to Stream that listens to WordPress AI Client `before`/`after_generate_result` hooks and writes audit records for each generation. Records include acting user, timestamp, provider, model, input/output/thinking token counts, duration, finish reason, and extended metadata. Full prompt and response text capture is supported via a **single opt-in** setting (`log_prompt_and_response_text`), disabled by default with a PII warning. Developer filters `wp_stream_ai_client_log_prompt` and `wp_stream_ai_client_log_response` allow redaction or omission of stored text. The Stream list table shows only the first line of multiline summaries so token/metadata rows stay scannable when prompt/response logging is enabled.

# Checklist

## Release Changelog

- New: AI Client connector logs WordPress AI Client generations (tokens, model, provider, duration) with optional full prompt/response capture behind a single opt-in toggle.

## Release Checklist

- [x] This pull request is to the `develop` branch.

## Test plan

- [ ] On WP 7.0+ with AI Client configured, trigger a generation and confirm a Stream record appears under **AI Client → Prompts → Generated** with user, model, provider, and token counts.
- [ ] With **Log Prompt and Response text** off (default), confirm prompt/response body are not stored in the record summary or meta.
- [ ] Enable **Log Prompt and Response text**, confirm the PII warning is visible, prompt/response appear in the record, and the list table shows only the first summary line.
- [ ] Hook `wp_stream_ai_client_log_prompt` / `wp_stream_ai_client_log_response` to redact text; confirm stored values reflect the filter output.
- [ ] On a site without `WP_AI_Client_Event_Dispatcher`, confirm Stream loads with no fatals and no AI records are written.
- [ ] Run PHPUnit: `vendor/bin/phpunit --filter Connector_AI_Client` and `vendor/bin/phpunit --filter Test_List_Table`.


readme.txt
```diff
@@ -139,6 +154,18 @@ Use only `$_SERVER['REMOTE_ADDR']` as the client IP address for event logs witho
...
+
+New Features:
+
+* Add AI Client connector for WordPress 7.0+: logs AI generation metadata (operation, provider, model, tokens, duration, and more) when the WordPress AI Client event dispatcher is available.
+
+Enhancements:
+
+* Show only the first line of multiline summaries in the Stream list table.
+
+[View the full release notes on GitHub.](https://github.com/xwp/stream/blob/master/changelog.md#unreleased)
```
bartoszgadomski: I think this entry shouldn't be included in readme.txt but only in ‎changelog.md‎ (which you already added).

connectors/class-connector-ai-client.php
```diff
@@ -0,0 +1,863 @@
...
+				'thought_tokens' => $thought_tokens,
+				'duration_ms'    => $duration_ms,
+				'prompt_text'    => $pending['prompt_text'],
+				'response_text'  => $response_text,
+				'finish_reason'  => $finish_reason,
+			);
+
+			$log_args = $this->append_result_context_to_log_args( $log_args, $event, $model, $result );
+			/* translators: 1: AI operation (e.g. "chat"), 2: provider slug, 3: model ID */
+			$message = esc_html__( '%1$s via %2$s/%3$s (tokens: %4$d/%6$d/%5$d) in %7$dms', 'stream' );
```
bartoszgadomski: **Token order in the summary looks swapped.**

`Log::log()` renders the summary with `vsprintf( $message, $args )`, so the placeholders map to `$log_args` by position: `%4$d` = `input_tokens`, `%5$d` = `output_tokens`, `%6$d` = `thought_tokens`. This template is `%4$d/%6$d/%5$d`, so the feed prints **input / thought / output**, while the args array, the meta keys, and the PR description all read input/output/thought.

If the order is deliberate, please document it; otherwise use `%4$d/%5$d/%6$d`. Nothing currently asserts the rendered summary, so a test on the formatted string would lock this down.

Two smaller notes on the same string:

- The `translators` comment only describes placeholders 1-3, but the template uses up to `%7$d` (and `%8$s`/`%9$s` below). Translators can't safely reorder what they can't identify.
- Other connectors use `__()` / `_x()` for log templates (e.g. `Connector_Menus`, `Connector_Posts`, `Connector_Jetpack`); `esc_html__()` here means any entity-worthy character in a *translation* gets stored HTML-encoded in the `summary` column. The value is already run through `wp_strip_all_tags()` on insert and `wp_kses()` on output, so the escaping is redundant.

connectors/class-connector-ai-client.php
```diff
@@ -0,0 +1,863 @@
...
+		);
+	}
+
+	/**
+	 * Removes unnecessary meta data from the record array.
+	 *
+	 * @param array<string, mixed> $record Record about to be inserted.
+	 * @return array<string, mixed>
+	 */
+	public function filter_wp_stream_record_array( array $record ) {
```
bartoszgadomski: **The `array` type hint can turn a filtered-out record into a fatal.**

`wp_stream_record_array` is not guaranteed to carry an array: `Connector_Woocommerce::callback_wp_stream_record_array()` returns `false` to drop a record (`connectors/class-connector-woocommerce.php:697`), and that callback runs on the same hook at the same priority 10.

Today this is safe only because `load_connectors()` instantiates `ai-client` before `woocommerce`, so this filter runs first. That ordering is not contractual — the `wp_stream_connectors` filter lets third parties reorder or inject connectors, and any other callback returning `false`/`null` before this one produces a `TypeError` on record insert rather than a skipped record.

Suggest dropping the hint and guarding instead:

```php
public function filter_wp_stream_record_array( $record ) {
	if ( ! is_array( $record ) ) {
		return $record;
	}
	// ...
}
```

Same reasoning applies to `add_settings_fields( array $fields )`.

connectors/class-connector-ai-client.php
```diff
@@ -0,0 +1,863 @@
...
+				'finish_reason'  => $finish_reason,
+			);
+
+			$log_args = $this->append_result_context_to_log_args( $log_args, $event, $model, $result );
+			/* translators: 1: AI operation (e.g. "chat"), 2: provider slug, 3: model ID */
+			$message = esc_html__( '%1$s via %2$s/%3$s (tokens: %4$d/%6$d/%5$d) in %7$dms', 'stream' );
+
+			// Add the prompt and response text to the message if enabled.
+			if ( $this->is_prompt_and_response_logging_enabled() ) {
+				$message .= sprintf( "\n\n[%s]\n%s\n\n[%s]\n%s", esc_html__( 'Prompt', 'stream' ), '%8$s', esc_html__( 'Response', 'stream' ), '%9$s' );
```
bartoszgadomski: **Worth calling out in the docs: prompt/response text reaches alert destinations.**

With the opt-in enabled, prompt and response bodies become part of `summary`, and Stream alerts forward `summary` verbatim to third parties: Slack (`alerts/class-alert-type-slack.php:146,150`), IFTTT (`alerts/class-alert-type-ifttt.php:282`), and the admin menu alert (`alerts/class-alert-type-menu-alert.php:51`).

So an admin who enables this setting also starts shipping AI prompt text to any configured webhook. The `readme.txt` PII warning covers storage but not egress. Either mention it in the warning copy, or note that `wp_stream_ai_client_log_prompt` / `wp_stream_ai_client_log_response` are the redaction point for that path too.

connectors/class-connector-ai-client.php
```diff
@@ -0,0 +1,863 @@
...
+	 * Supports AbstractEnum (WP AI Client), BackedEnum, UnitEnum, and objects
+	 * with a ->value property. Unknown values pass through as the operation label;
+	 * null or unresolvable capabilities fall back to 'chat'.
+	 *
+	 * @param object|null $capability CapabilityEnum instance or null.
+	 * @return string
+	 */
+	private function normalize_capability( $capability ) {
+		if ( null === $capability ) {
+			return 'unknown';
```
bartoszgadomski, line 697: **Docblock contradicts the code, and there are two "unknown" slugs.**

The docblock above says null or unresolvable capabilities "fall back to `'chat'`", but this returns `'unknown'` for null and `'unknown_operation'` for an unresolvable value.

Since `operation` is user-facing in the summary and filterable in the list table, two distinct slugs for what is effectively the same "couldn't determine it" case will fragment filtering. Suggest a single value and a docblock that matches.

connectors/class-connector-ai-client.php
```diff
@@ -0,0 +1,863 @@
...
+		if ( is_string( $value ) ) {
+			return $value;
+		}
+
+		if ( ! is_object( $value ) ) {
+			return is_scalar( $value ) ? (string) $value : '';
+		}
+
+		if ( $value instanceof \WordPress\AiClient\Common\AbstractEnum ) {
+			return isset( $value->value ) ? (string) $value->value : (string) $value->name;
```
bartoszgadomski: **This branch always resolves to `->name`, by the method's own reasoning.**

The docblock states that `AbstractEnum` has `__get()` but no `__isset()`, so `isset( $value->value )` is always `false`. That makes this line equivalent to `return (string) $value->name;`, which yields the enum *name* (e.g. `TEXT_GENERATION`) rather than its *value* (e.g. `text_generation`), and skips the `__toString()` path below that the docblock describes as returning the value.

It happens to come out right for capabilities because `normalize_capability()` lowercases, and for roles because `prompt_section_key_for_scalar_role()` lowercases — but any enum whose value isn't just the lowercased name (hyphens, abbreviations) would produce a wrong slug. Using `__toString()` for `AbstractEnum`, or `property_exists()` / a `try` around `->value`, would express the intent directly.

connectors/class-connector-ai-client.php
```diff
@@ -0,0 +1,863 @@
...
+
+		// Total and thought tokens.
+		try {
+			$tu = $result->getTokenUsage();
+			if ( method_exists( $tu, 'getTotalTokens' ) ) {
+				$log_args['total_tokens'] = (int) $tu->getTotalTokens();
+			}
+			if ( method_exists( $tu, 'getThoughtTokens' ) ) {
+				$thought_tokens             = $tu->getThoughtTokens();
+				$log_args['thought_tokens'] = null === $thought_tokens ? null : (int) $thought_tokens;
```
bartoszgadomski: Minor: setting `thought_tokens` to `null` here makes the two outputs disagree. `Log::log()` strips null values before writing meta, so the meta key disappears, while `vsprintf()` still receives `null` for `%6$d` and prints `0`. The test doubles return `null` from `getThoughtTokens()`, so this is the default path, not an edge case.

Also, `thought_tokens` is already computed at line 314 with `(int)`; this recomputes it. Keeping the `(int)` cast (0 when unsupported) or omitting the key entirely would make summary and meta consistent.

connectors/class-connector-ai-client.php
```diff
@@ -0,0 +1,863 @@
...
+		}
+
+		$system_block = "\n[System]\n" . $model_instruction;
+
+		if ( '' === $message_prompt ) {
+			return $system_block;
+		}
+
+		// Already has section headings — prepend system block directly.
+		if ( false !== strpos( $message_prompt, '--- ' ) ) {
```
bartoszgadomski: Two cosmetic issues in the assembled prompt text:

- Heading styles don't match. `extract_prompt_text()` emits `--- User ---`, while this method emits `[System]` and `[User]`. A single reader sees both conventions in one record.
- This heuristic keys off the literal `'--- '`, so a prompt whose body contains a Markdown horizontal rule, a diff hunk, or any `--- ` sequence is treated as already-sectioned and loses its `[User]` heading.

Passing the section list (or a boolean) down from `extract_prompt_text()` instead of re-detecting via `strpos()` would remove the guesswork. Display-only, no data loss.

connectors/class-connector-ai-client.php
```diff
@@ -0,0 +1,863 @@
...
+				 * @param string $prompt_text The prompt text about to be logged.
+				 * @param object $event       The BeforeGenerateResultEvent instance.
+				 */
+				$prompt_text = (string) apply_filters( 'wp_stream_ai_client_log_prompt', $prompt_text, $event );
+			}
+
+			// Store the pending data in the pending array, so we can match the after-hook.
+			$pending = $this->get_pending_storage();
+			$this->evict_oldest_pending_if_full( $pending );
+			$pending[ $model ] = array(
```
bartoszgadomski, line 258: Low priority / possibly by design: keying on the model instance pairs before/after correctly for sequential calls and for concurrent calls on *distinct* instances (both covered by tests), but two overlapping generations on the *same* instance overwrite this entry.

The realistic trigger is nesting — a generation started from inside an `after_generate_result` listener, or a helper that reuses one cached model across a nested call. The result is one lost record plus a wrong duration and mismatched prompt on the other.

If WP AI Client can't nest on one instance, a note here saying so would close the question; otherwise a small stack per model would cover it.

classes/class-list-table.php
```diff
@@ -430,6 +430,27 @@ public function column_default( $item, $column_name ) {
 		echo wp_kses( $out, $allowed_tags );
 	}
 
+	/**
+	 * Formats a summary for the admin list table: first line only.
+	 *
+	 * @param string $summary Full summary stored in the database.
+	 * @return string Summary preview text.
+	 */
+	public static function format_summary_preview_html( $summary ) {
```
bartoszgadomski: **This changes the summary column for every connector, not just AI Client.**

`column_default()` routes all records through this, so any multiline summary from any connector (including third-party ones) is now truncated in the activity feed. The changelog frames it as a general enhancement, which is fair, but it's a behavior change beyond the scope suggested by the PR title — is that intended?

Two smaller points:

- There is no affordance signalling that content was hidden. A reader sees a complete-looking line with no hint to open the record. A trailing `…` would help.
- The name says `_html` while the return value is plain text (the caller still passes it through `wp_kses()`). For a new `public static` method, `format_summary_preview()` would describe it more accurately.

bartoszgadomski (changes requested):
Nice work @shadyvb! I left inline comments, please take a look.

connectors/class-connector-ai-client.php
shadyvb: The order is deliberate, yes. Fixed the escaping and translator comment as well.

connectors/class-connector-ai-client.php
shadyvb: WP AI Client runs a strict before → execute → after sandwich. Same-instance reuse is sequential, not nested.

readme.txt
```diff
@@ -101,6 +102,20 @@ As a workaround, you can use the `wp_stream_client_ip_address` filter to adapt t
 ⚠️ **WARNING:** The above is an insecure workaround that you should only use when you fully understand what this implies. Relying on any variable with the `HTTP_*` prefix is prone to spoofing and cannot be trusted!
 
 
+= AI Client (WordPress 7.0+) =
+
+When WordPress 7.0+ provides the AI Client event dispatcher (`WP_AI_Client_Event_Dispatcher`), Stream logs each AI generation as an activity record under **AI Client → Prompts → Generated**. If the dispatcher is not available, the connector registers no hooks and has no effect.
+
+By default, Stream stores metadata for every generation: operation, provider, model, input/output/thought token counts, duration, finish reason, and other extended fields when present. The activity summary shows a one-line preview (for example, `chat via openai/gpt-4o (tokens: 120/0/45) in 842ms`). Multiline summaries display only the first line in the list table; open a record to see the full text.
```
bartoszgadomski: Leftover docs: this still describes the counts as **input/output/thought**, but the summary template (and the PHPUnit assertion) renders **input/thought/output**. The example `120/0/45` already matches that display order.

Please align the prose with the deliberate `%4$d/%6$d/%5$d` format, e.g. “input/thought/output token counts” and a short note that the three numbers in the preview are input, thought, then output.

readme.txt
```diff
@@ -101,6 +102,20 @@ As a workaround, you can use the `wp_stream_client_ip_address` filter to adapt t
...
 
 
+= AI Client (WordPress 7.0+) =
+
+When WordPress 7.0+ provides the AI Client event dispatcher (`WP_AI_Client_Event_Dispatcher`), Stream logs each AI generation as an activity record under **AI Client → Prompts → Generated**. If the dispatcher is not available, the connector registers no hooks and has no effect.
+
+By default, Stream stores metadata for every generation: operation, provider, model, input/output/thought token counts, duration, finish reason, and other extended fields when present. The activity summary shows a one-line preview (for example, `chat via openai/gpt-4o (tokens: 120/0/45) in 842ms`). Multiline summaries display only the first line in the list table; open a record to see the full text.
+
+Prompt and response text are **not** logged by default. To opt in, enable **Log Prompt and Response text** under **Stream → Settings → AI Client**. When enabled, prompt and response text are appended to the activity summary (not stored in meta, which is size-limited). **Privacy Warning:** This content may include personally identifiable information (PII). Ensure your privacy policy covers AI data collection before enabling.
```
bartoszgadomski: Leftover docs: the Settings checkbox warning now covers Slack/IFTTT (and the redaction filters), but this `readme.txt` PII paragraph still only talks about storage.

Please mirror the settings copy here so operators who never open Stream → Settings still see that prompt/response text may be forwarded to configured alerts or webhooks, and that `wp_stream_ai_client_log_prompt` / `wp_stream_ai_client_log_response` are the redaction point.

connectors/class-connector-ai-client.php
```diff
@@ -0,0 +1,866 @@
...
+		// do_action( 'wp_ai_client_{event}' ). Core wires it in wp-settings.php
+		// before init, so this is true at connector load time on WP 7.0+.
+		return class_exists( 'WP_AI_Client_Event_Dispatcher' );
+	}
+
+	/**
+	 * Injects the AI Client settings section into Stream's settings fields array.
+	 *
+	 * Adds opt-in checkbox — log_prompt_and_response_text — under
+	 * a dedicated "AI Client" section in Stream → Settings. Both default to off.
```
bartoszgadomski: Leftover docs: this still says “Both default to off”, but there is only one checkbox (`log_prompt_and_response_text`). Looks like a leftover from an earlier two-toggle design — please drop “Both”.

connectors/class-connector-ai-client.php
```diff
@@ -0,0 +1,866 @@
...
+	public function add_settings_fields( $fields ) {
+		if ( ! is_array( $fields ) ) {
+			return $fields;
+		}
+
+		$pii_warning = sprintf(
+			'<strong>%s</strong> %s %s',
+			esc_html__( 'Privacy Warning:', 'stream' ),
+			esc_html__( 'This content may include personally identifiable information (PII). Ensure your privacy policy covers AI data collection before enabling.', 'stream' ),
+			esc_html__( 'When enabled, prompt and response text are stored in the record summary and may be forwarded verbatim to any configured Stream alerts or webhooks (e.g. Slack, IFTTT). Use the wp_stream_ai_client_log_prompt and wp_stream_ai_client_log_response filters to redact or omit text before it is stored.', 'stream' )
```
bartoszgadomski: **“Verbatim” overstates what Stream actually persists.**

With this setting on, prompt/response text is written into `summary` via `log()`, then `DB::sanitize_record()` runs `wp_strip_all_tags()` on every scalar field before insert (`classes/class-db.php`). Markup, `List<T>`-style generics, and anything that looks like an HTML tag is altered or removed, so the stored (and alert-forwarded) text is not a byte-for-byte copy of the generation.

That sanitizer is existing Stream behavior, not new in this PR — but this connector is the first to put user-authored bodies in `summary`, and this warning is the first to promise they are stored and forwarded verbatim.

Please either soften the wording (e.g. that HTML is stripped on insert) or document the limitation next to the opt-in. Same note belongs in `readme.txt` if you add the alerts sentence there.

bartoszgadomski (commented):
Follow-up notes after the review-response commits.

readme.txt
shadyvb: Fixed — readme now describes input/thought/output token counts and notes that the three preview numbers are input, thought, then output.

readme.txt
shadyvb: Fixed — the readme PII paragraph now covers alerts/webhooks forwarding, HTML stripping on insert, and the wp_stream_ai_client_log_prompt / wp_stream_ai_client_log_response redaction filters.

connectors/class-connector-ai-client.php
shadyvb: Fixed — dropped the leftover “Both” so the docblock matches the single checkbox.

connectors/class-connector-ai-client.php
shadyvb: Fixed — the settings warning no longer says “verbatim”; it now notes that HTML is stripped on insert, and the same limitation is documented in readme.txt.

bartoszgadomski (approved):
Thank you for addressing the feedback, @shadyvb!

PR #1974, opened 2026-08-31
Upgrade PHPUnit to 11.x and reconcile with host unit-test tier (XWPENG-43)

shadyvb:
Fixes XWPENG-43.

Base branch: `ticket/XWPENG-42-unit-test-tier` (host PHPUnit unit suite from XWPENG-42). This PR upgrades the test toolchain to PHPUnit 11 and reconciles integration/unit test discovery with that foundation.

## Summary

- **PHPUnit 9.6 → 11.5** with Solution #3 shim for `expectDeprecated()` on `WP_StreamTestCase` (PHPUnit 11 removed it; trait restores deprecated-expectation behavior where needed).
- **Reconciled with XWPENG-42:** `phpunit-unit.xml` on PHPUnit 11 schema; `Connectors_Unit_Test` / `Query_Unit_Test` updated with `#[DataProvider]` attributes.
- **Test discovery:** `phpunit.xml` / integration configs use `suffix="Test.php"`; ~58 integration test files renamed; removed `WP_Stream_` prefix from class names (ability tests live in namespace `WP_Stream`, e.g. `Ability_Create_Alert_Test`).
- **Connector integration tests:** removed `Mock_Call_Recorder`; expectations use split `once()->with()` (and named static `record_log_call` callbacks where multiple log entries must be matched in order).
- **Parity fixes (post-review):** ACF `save_post` expectations; Mercator `make_primary()`; Users `get_password_reset_key()` multi-log scenarios via per-class static `record_log_call` helpers.
- **Multisite fix:** `ensure_mercator_mapping_table()` in testcase—Admin ajax tests left stale `mercator.db.version` without creating the mapping table, polluting Mercator connector tests.
- **Cleanup:** removed dead `wp_stream_preload_coverage_sources()` and `tests/phpunit/class-mock-call-recorder.php`.
- **Tooling:** Composer `config.platform` 8.2, Docker PHP 8.2 default, `switch-to:php7.4` build args, `contributing.md` corrections.
- **`Filter_Input_Test`:** `expectException()` + `#[WithoutErrorHandler]` on the regex validation test.
- **`CLI_Test`:** renamed from `test-class-cli.php` for PHPUnit 11 discovery; connection-check coverage uses a named filter callback instead of a closure.

### Commits on this PR (`ticket/XWPENG-42-unit-test-tier..HEAD`)

| Commit | Description |
|--------|-------------|
| `8a97c517` | **XWPENG-43** — PHPUnit 11 migration, test renames, connector expectation refactor, tooling, parity fixes. |
| `20db0252` | Fix PHPCS violations in PHPUnit connector tests. |
| `19d9307a` | Rename CLI tests for PHPUnit 11 `Test.php` discovery; named filter callback. |

### Breaking / acceptance criteria notes

- Jira AC amended: **PHP 8.2+** required for the test toolchain (not 8.0/8.1).
- `config.platform` 8.2; plugin **Requires PHP** remains 7.2 on this branch (raised separately in XWPENG-46).
- Integration suites: relaxed `failOnWarning` / `failOnNotice`; scoped `WithoutErrorHandler` where WordPress emits expected noise.
- No CI “423 guard” (declined for this effort).

### Gotchas

- Polyfills alias PHPUnit 4.0.0 as 1.1.5 for `wp-test-utils` compatibility.
- `@group` annotations remain on integration tests.
- EDD connector anonymous closure is pre-existing and unchanged.

## Test plan

- [ ] `composer test-unit` — 53 tests, exit 0
- [ ] `npm run test` — 53 unit + 423 single-site + 423 multisite integration, all exit 0
- [ ] `composer lint` and lint-tests pass
- [ ] Spot-check connector tests (ACF, Mercator, Users) for ordered multi-log expectations after Mock_Call_Recorder removal

# Checklist

## Release Changelog

- Fix: N/A (developer tooling / test infrastructure).
- New: N/A.

## Release Checklist


tests/phpunit/connectors/Connector_Editor_Test.php
```diff
@@ -44,61 +42,36 @@ public function tearDown(): void {
 		file_put_contents( WP_PLUGIN_DIR . '/hello.php', $this->original_contents ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_operations_file_put_contents
```
bartoszgadomski, line 42: `tearDown()` overrides the polyfills' `tearDown()` but never calls `parent::tearDown()`, so `WP_Ajax_UnitTestCase::tear_down()` and `WP_UnitTestCase_Base::tear_down()` never run for this class. For both tests here that means:

- no `ROLLBACK` — the next test's `start_transaction()` issues `START TRANSACTION`, which implicitly commits whatever these tests wrote (the admin user, for one);
- `_restore_hooks()` is skipped, so the mocked connector stays registered on `wp_ajax_edit-theme-plugin-file`;
- `wp_set_current_user( 0 )`, the `$_POST`/`$_GET` reset and `set_current_screen( 'front' )` are all skipped;
- `error_reporting()` keeps the `& ~E_WARNING` mask that `WP_Ajax_UnitTestCase::set_up()` applies. Every later test saves that already-masked value in its own `set_up()` and restores it, so the mask sticks for the rest of the process.

It does pass today — PHPUnit drops the invocation mocker after each test in `__phpunit_verify()`, so the leftover mock's `log()` is an inert stub rather than a second `once()` violation — and the missing call predates this PR. But the class now has two tests sharing that state, so this is a good moment to fix it:

```suggestion
		file_put_contents( WP_PLUGIN_DIR . '/hello.php', $this->original_contents ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_operations_file_put_contents
		parent::tearDown();
```

tests/phpunit/connectors/Connector_Settings_Test.php
```diff
@@ -42,118 +42,160 @@ public function test_is_option_ignored() {
...
+	}
+
+	public function test_callback_updated_option_permalink_structure() {
+		$add_method    = is_multisite() ? 'add_site_option' : 'add_option';
+		$update_method = is_multisite() ? 'update_site_option' : 'update_option';
+
+		call_user_func( $add_method, 'permalink_structure', '' );
+		$this->simulate_customize_save();
+
+		$this->mock->expects( $this->atLeastOnce() )
```
bartoszgadomski, line 107: This test and the `category_base` / `tag_base` ones (lines 141 and 175) relax the count to `atLeastOnce()`, while `test_callback_updated_option_users_can_register` keeps `once()`. The combined test these were split from asserted `exactly( 4 )` across four option writes, i.e. exactly one record per option, so `once()` should still hold here and would keep catching duplicate audit records.

If a single permalink/category/tag write genuinely logs more than once (a rewrite flush writing the option twice, say), a one-line comment saying so would help. Otherwise `once()` preserves the original contract.

phpunit.xml
```diff
@@ -4,37 +4,39 @@
 	bootstrap="tests/bootstrap.php"
 	backupGlobals="false"
 	colors="true"
-	verbose="true"
-	convertErrorsToExceptions="true"
-	convertNoticesToExceptions="true"
-	convertWarningsToExceptions="true"
-	xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.3/phpunit.xsd"
+	cacheDirectory=".phpunit.cache"
+	failOnWarning="false"
```
bartoszgadomski, line 8: The same run reports `PHPUnit Deprecations: 87` and nothing surfaces them. `displayDetailsOnPhpunitDeprecations="true"` would make them visible. Most are probably the remaining doc-comment metadata — `tests/` still has 23 `@group` and 7 `@requires` — and annotation metadata is removed in PHPUnit 12, so this is the tail of the same migration. The 7 `@requires PHPUnit 5.7` are stale either way and can simply be deleted.

tests/testcase.php
```diff
@@ -32,6 +32,64 @@ public function setUp(): void {
...
+	 * which was removed in PHPUnit 10. We still register the doing-it-wrong /
+	 * deprecated catchers so setExpectedIncorrectUsage() / setExpectedDeprecated()
+	 * keep working. Annotation parsing is omitted (zero @expectedDeprecated uses;
+	 * the one @expectedIncorrectUsage was converted).
+	 *
+	 * @link https://core.trac.wordpress.org/ticket/59486#comment:8
+	 * @link https://core.trac.wordpress.org/ticket/62004
+	 * @return void
+	 */
+	public function expectDeprecated(): void { // phpcs:ignore WordPress.NamingConventions.ValidFunctionName.MethodNameInvalid
```
bartoszgadomski, line 79: The shim looks right and the docblock is honest about omitting annotation parsing. The residual risk is that it fails silently: `@expectedDeprecated` / `@expectedIncorrectUsage` on a future test will now be ignored rather than erroring, and the next person has no way to notice the annotation is inert.

Worth a line in `contributing.md` (or here) pointing at `setExpectedDeprecated()` / `setExpectedIncorrectUsage()` as the only supported form.

tests/phpunit/connectors/Connector_Users_Test.php
```diff
@@ -7,130 +7,147 @@
...
+				$this->greaterThan( 0 )
 			);
 
-		// Do stuff.
+		self::factory()->user->create( array( 'display_name' => 'TestGuy' ) );
+
+		$this->assertFalse( 0 === did_action( $this->action_prefix . 'callback_user_register' ) );
+	}
+
+	public function test_callback_user_register_by_logged_in_user() {
```
bartoszgadomski, line 67: This test creates `TestGuy` before `expects()` is configured, so the first `New user registration` call reaches the mock while it still has no matcher and only the second create is counted. That is what makes `once()` correct here, but the ordering is load-bearing and unmarked — moving `expects()` above the first `create()` breaks the test for a non-obvious reason.

A short comment noting that the first registration is intentionally unobserved would do it, or set the current user in `setUp()` so the test body only performs the observed create.

tests/phpunit/connectors/Connector_ACF_Test.php
```diff
@@ -9,25 +9,44 @@
...
+	 */
+	private static $recorded_log_calls = array();
+
+	/**
+	 * Records log() arguments for post-hoc assertions.
+	 *
+	 * @param mixed ...$args Log method arguments.
+	 * @return void
+	 */
+	public static function record_log_call( ...$args ) {
```
bartoszgadomski, line 33: `$recorded_log_calls` and `record_log_call()` are duplicated verbatim in `Connector_Users_Test` and `Connector_Mercator_Test`. Since `WP_StreamTestCase` already gained shared helpers in this PR, a `protected static` recorder there (with the reset in its `setUp()`) would remove three copies and give the ordered-capture pattern one documented home.

tests/phpunit/connectors/Connector_Posts_Test.php
```diff
@@ -0,0 +1,706 @@
...
+			)
+		);
+	}
+
+	/**
+	 * Expected post transition context subset.
+	 *
+	 * @var array
+	 */
+	private static $expected_post_context = array();
```
bartoszgadomski, line 625: Nit: this property is declared after all the test methods, ~600 lines below the top of the class (same pattern in `Connector_Settings_Test` line 283 and in `Connector_Media_Test`), whereas the rest of the suite declares properties at the top with a docblock.

It's also not reset in `setUp()`, so a test that forgets to assign it would validate against the previous test's expectations instead of failing. Resetting it in `setUp()` would turn that into a hard failure.

bartoszgadomski (commented):

PR #1977, opened 2026-09-01
Expand Playwright E2E, migrate to wp-env, PHP 8.2 minimum, and modernize internals (XWPENG-44–47)

shadyvb:
Fixes XWPENG-44, XWPENG-45, XWPENG-46, XWPENG-47.

Base branch: `ticket/XWPENG-43-phpunit-11` (#1974). This PR stacks on the PHPUnit 11 migration and delivers local-dev migration, expanded E2E coverage, the PHP 8.2 minimum / 5.0.0 release alignment, Settings/DB hardening, and a Rector-driven modernization pass.

### XWPENG-45 — wp-env local dev & CI

- Replaces Docker Compose with `@wordpress/env` (`.wp-env.json`): multisite, Email Logger mu-plugin, `lifecycleScripts.afterStart` network activation, and `env.tests.mappings` for PHPUnit helper plugins.
- Removes legacy Docker assets (`docker-compose*.yml`, `local/docker/`, `local/public/` WordPress bootstrap) and the `docker-images.yml` workflow.
- Updates `lint-and-test.yml`, deploy workflows, `contributing.md`, `tests/bootstrap.php`, and `tests/wp-tests-config-wp-env.php` for wp-env.
- Adds `switch-to:php8.2|8.3|8.4` npm scripts and documents MCP / Application Passwords on localhost.

### XWPENG-44 — Playwright E2E

- Adds five specs: `alert-create`, `export-download`, `network-activated-records`, `records-filter`, `settings-save`.
- Adds shared helper `tests/e2e/helpers/stream-plugin.js`; updates existing specs and `playwright.config.js` for wp-env URLs and setup.

### XWPENG-46 — PHP 8.2 minimum & 5.0.0 release

- Raises minimum PHP from 7.2 to **8.2** with an admin notice (no fatal) for unsupported runtimes; bumps plugin version to **5.0.0** in `stream.php`, `class-plugin.php`, and `readme.txt`.
- Aligns Composer `require.php`, PHPCS `testVersion`, CI matrix (8.2 / 8.3 / 8.4), and `changelog.md`.
- Adds `Fail_Php_Version_Unit_Test` and `Connectors_Unit_Test` coverage for the version gate.

### XWPENG-47 — Settings filter boundaries & DB driver contract

- Guards `wp_stream_settings_option_key`, `wp_stream_settings_options`, and `wp_stream_settings_option_fields` so non-string / non-array filter returns fall back safely before typed-property assignment.
- Declares `DB_Driver::setup_storage(): Install` on the interface and WPDB implementation (breaking for third-party drivers omitting the return type).

### Codebase modernization (Rector 2.6)

- Adds `rector.php` and `composer lint-rector` / `composer rector` scripts.
- Applies typed properties across `classes/`, connectors, alerts, and exporters; constructor property promotion; two `switch` → `match` conversions.
- Extracts `Plugin::boot()` from the constructor.
- Splits `Admin` into focused collaborators (`Admin_Menu`, `Admin_Assets`, `Admin_Screen_Records`, `Admin_Screen_Settings`, `Admin_Ajax`, `Admin_Purge`) with colocated hook registration and dedicated PHPUnit coverage.

### Commits on this PR (`ticket/XWPENG-43-phpunit-11..HEAD`)

| Commit | Description |
|--------|-------------|
| `ae576e71` | **XWPENG-44** — Expand Playwright E2E coverage for Stream admin flows. |
| `1a4675dc` | **XWPENG-45** — Migrate local dev and test tooling to wp-env. |
| `0c21fd72` | **XWPENG-46** — Raise minimum PHP to 8.2; matrix CI for supported runtimes. |
| `0956d649` | **XWPENG-46** — Align release to 5.0.0; PHP version gate unit tests. |
| `2c0932e5` | Add Rector 2.6 dev tooling and empty `rector.php`. |
| `47ed6f88` | Apply typed properties in `classes/` via Rector. |
| `ea122a51` | **XWPENG-47** — Harden Settings filter boundaries; tighten `DB_Driver::setup_storage` contract. |
| `49cde6ba` | Type the Connector family (22 children). |
| `3a0f708c` | Type Alert_Type, Alert_Trigger, and Exporter families. |
| `77ce1e09` | Apply constructor property promotion via Rector. |
| `3ad23fc1` | Convert two switch statements to match expressions. |
| `13f24b07` | Extract `Plugin::boot()` from the constructor. |
| `251a8f98` | Split Admin god-class into focused collaborator classes. |
| `ef7ac652` | Fix PHPCS formatting in Admin collaborator tests. |

## Test plan

- [x] 53 unit tests pass (`composer test-unit`)
- [x] 472 integration tests pass ×2 (single-site + multisite)
- [x] 27 E2E tests pass (Playwright)
- [x] Lint pass (`npm run lint`)
- [ ] Reviewer: `npm run wp-env start` and smoke-test admin URLs locally
- [ ] Reviewer: confirm stacked merge order after #1974 lands

# Checklist

## Release Changelog

- Fix: Guard Settings filter callbacks so non-array / non-string returns cannot cause TypeErrors when typed properties are assigned ([XWPENG-47](https://xwp-co.atlassian.net/browse/XWPENG-47)).
- New: N/A (see breaking change below).
- **Breaking:** Minimum PHP raised to 8.2; plugin version **5.0.0** ([XWPENG-46](https://xwp-co.atlassian.net/browse/XWPENG-46)). Third-party `DB_Driver` implementations must declare `setup_storage(): Install` ([XWPENG-47](https://xwp-co.atlassian.net/browse/XWPENG-47)).

## Release Checklist

- [x] Release version follows [semantic versioning](https://semver.org). Does it include breaking changes? **Yes — 5.0.0, PHP 8.2 minimum.**
- [x] Update changelog in `readme.txt`.
- [x] Bump version in `stream.php`.
- [x] Bump `Stable tag` in `readme.txt`.
- [x] Bump version in `classes/class-plugin.php`.

[XWPENG-47]: https://xwp-co.atlassian.net/browse/XWPENG-47?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


playwright.config.js
```diff
@@ -10,6 +10,30 @@ const { defineConfig, devices } = require( '@playwright/test' );
...
+	...devices[ 'Desktop Chrome' ],
+	storageState: 'playwright/.auth/user.json',
+};
+
+/**
+ * Specs that save Settings → General (options-general.php) to seed a Stream
+ * record. Concurrent saves race on the success notice, so they run in the
+ * serial `settings-ui` project instead of `chromium`.
+ */
+const settingsUiSpecs = [
```
bartoszgadomski, line 26: **Race: `settings-save.spec.js` is not isolated.**

It is not in `settingsUiSpecs` and it is not `networkActivatedSpec`, so it is not covered by the `chromium` `testIgnore` on line 68 and runs in `chromium` with `fullyParallel: true` (2 workers on CI, 4 locally) — while it mutates *network-wide* Stream settings (records TTL, cron tracking, exclude IP).

Concretely: `network-activated-records.spec.js` also runs in `chromium`, and its `beforeEach` calls `clearNetworkExcludeRulesViaWpCli()` with no argument. The helper's default parameter is `'203.0.113.44'` — the same IP `settings-save` saves and then asserts on after reload. A concurrent worker can delete that row between the save and the assertion.

Suggest adding `'**/settings-save.spec.js'` to `settingsUiSpecs` (it saves shared settings, which is exactly the reason that project exists), and changing the no-arg helper default to clear only empty rows so global cleanup cannot delete a rule another spec owns.

phpunit.xml
```diff
@@ -17,7 +17,7 @@
 		</report>
 	</coverage>
 	<php>
-		<const name="WP_TEST_ACTIVATED_PLUGINS" value="advanced-custom-fields/acf.php,easy-digital-downloads/easy-digital-downloads.php,jetpack/jetpack.php,user-switching/user-switching.php,wordpress-seo/wp-seo.php,two-factor/two-factor.php"/>
+		<const name="WP_TEST_ACTIVATED_PLUGINS" value="advanced-custom-fields/acf.php,easy-digital-downloads/easy-digital-downloads.php,jetpack/jetpack.php,user-switching/user-switching.php,wordpress-seo/wp-seo.php,two-factor/two-factor.php,wp-email-logger/wp-email-logger.php"/>
```
bartoszgadomski, line 20: Two things about force-activating `wp-email-logger` in the PHPUnit suites:

1. **It generates Stream records.** On multisite the plugin stores its log via `update_site_option( 'network_emails_log', … )`. Stream's Settings connector registers `update_site_option`, and `callback_updated_option()` logs *any* option not in the small `is_option_ignored()` list (`_transient_*`, `_site_transient_*`, trailing `$`, and three image/size options) — `network_emails_log` isn't excluded. So every mail sent during the multisite suite also writes a `settings`/`updated` record whose value is the flattened email log. The suite passes today, but it is hidden coupling for any record count/ordering assertion, plus larger record meta. (Single-site is unaffected: `callback_update_option()` only logs for WP-CLI/customizer and only for options in `$labels`.)
2. **Is it needed for PHPUnit at all?** `alert-create.spec.js` deliberately uses the Highlight alert type rather than asserting via Email Logger, so this dependency reads as E2E/manual-testing oriented. Leaving it out of `WP_TEST_ACTIVATED_PLUGINS` would avoid both the record noise and making the integration suite depend on the `make.wordpress.org` zip being reachable.

For the record, the zip layout is fine — it contains `wp-email-logger/wp-email-logger.php`, so the path used here is correct.

phpunit-multisite.xml
```diff
@@ -17,7 +17,7 @@
 		</report>
 	</coverage>
 	<php>
-		<const name="WP_TEST_ACTIVATED_PLUGINS" value="advanced-custom-fields/acf.php,easy-digital-downloads/easy-digital-downloads.php,jetpack/jetpack.php,user-switching/user-switching.php,two-factor/two-factor.php"/>
+		<const name="WP_TEST_ACTIVATED_PLUGINS" value="advanced-custom-fields/acf.php,easy-digital-downloads/easy-digital-downloads.php,jetpack/jetpack.php,user-switching/user-switching.php,wordpress-seo/wp-seo.php,two-factor/two-factor.php,wp-email-logger/wp-email-logger.php"/>
```
bartoszgadomski, line 20: This also adds `wordpress-seo/wp-seo.php`, which was previously single-site only (it was in `phpunit.xml` but not here). `tests/bootstrap.php` still contains:

```php
if ( ! is_multisite() ) {
	activate_plugin( 'wordpress-seo/wp-seo.php' );
}
```

Since the constant is applied through the `site_option_active_sitewide_plugins` filter, Yoast is now network-active for the whole multisite suite while the bootstrap still deliberately skips activating it there — the two now disagree.

Intentional? If yes, the bootstrap guard is misleading and should go (or get a comment); if it was just config alignment, this quietly changes what loads for every multisite test.

.wp-env.json
```diff
@@ -0,0 +1,30 @@
...
+  "phpVersion": "8.2",
+  "plugins": [".", "https://make.wordpress.org/test/files/2026/02/wp-email-logger.zip"],
+  "multisite": true,
+  "config": {
+    "WP_DEBUG": true,
+    "WP_DEBUG_LOG": true,
+    "SCRIPT_DEBUG": true
+  },
+  "lifecycleScripts": {
+    "afterStart": "wp-env run cli -- wp plugin activate stream wp-email-logger --network"
```
bartoszgadomski, line 11: `wp plugin activate stream` assumes the plugin is mounted at `wp-content/plugins/stream`, but wp-env derives that name from the checkout directory, not the plugin slug. For a local source like `"."`, `packages/env/lib/config/parse-source-string.js` does `const basename = path.basename( sourcePath )`.

So a clone into `stream-plugin/`, `stream-wp/`, `stream-2/`, etc. mounts at `wp-content/plugins/<dir name>` and then:

- this `afterStart` script fails (`stream` is not installed),
- `--env-cwd=wp-content/plugins/stream` in the `test:*` scripts fails,
- the `launch.json` path mapping silently stops matching.

CI is unaffected (checkout dir = repo name), so this is a contributor sharp edge. Either resolve the directory name dynamically or document in contributing.md that the clone must be named `stream`.

package.json
```diff
@@ -46,30 +47,27 @@
     "format:php": "composer format",
     "format:js": "npm run lint:js -- --fix",
     "test": "npm-run-all test:*",
-    "test:php-unit": "npm run cli -- composer test-unit --working-dir=wp-content/plugins/stream-src",
-    "test:php": "npm run cli -- composer test --working-dir=wp-content/plugins/stream-src",
-    "test:php-multisite": "npm run cli -- composer test-multisite --working-dir=wp-content/plugins/stream-src",
+    "test:php-unit": "wp-env start && wp-env run tests-cli --env-cwd=wp-content/plugins/stream -- composer test-unit",
+    "test:php": "wp-env start && wp-env run tests-cli --env-cwd=wp-content/plugins/stream -- vendor/bin/phpunit --no-coverage",
```
bartoszgadomski, line 51: These now bypass the Composer scripts rather than delegating to them, which changes behavior beyond the container swap:

- `composer test` is `phpunit --coverage-text` + `make-clover-relative.php`. With `--no-coverage` and no post-step, `tests/reports/clover.xml` is never written — so `test-report` (`php-coveralls`, line 57) has nothing to upload, and the coveralls badge in readme.md can't be refreshed from this path.
- `test:php-multisite` (line 52) re-declares inline what `composer test-multisite` already owns, so the multisite suite config now lives in two places.
- `test-xdebug` (line 55) and `test-one` (line 56) each run a single suite, while their Composer counterparts run single-site **and** multisite.

Delegating to `composer test` / `test-multisite` / `test-xdebug` / `test-one` inside `wp-env run tests-cli` keeps one source of truth — `test:php-unit` on line 50 already does exactly that.

Minor: every `test:*` script prefixes `wp-env start`, and CI already starts the env in its own step, so a full `npm run test` boots wp-env four times. Probably part of why the job timeout went 10 → 20.

tests/e2e/helpers/stream-plugin.js
```diff
@@ -0,0 +1,383 @@
...
+
+/**
+ * Network-activate Stream via wp-cli (`stream` in wp-env).
+ *
+ * Use when the spec must wait until connectors are loaded (e.g. before a
+ * wp-cli seed). The plugins-row click can finish before Stream is actually
+ * active; this call is synchronous.
+ */
+export function networkActivateStreamViaWpCli() {
+	const slugs = [ 'stream', 'stream-src' ];
```
bartoszgadomski, line 115: `networkActivateStreamViaWpCli()` has no call sites in this PR (only the definition), and the `stream-src` fallback is the Compose-era mount name that no longer exists under wp-env — `plugins: ["."]` mounts at the checkout directory name, never `stream-src`. So the fallback can only ever fail before rethrowing.

Either drop the function, or if it is kept for upcoming specs, derive the slug from the actual mount name rather than hardcoding two guesses.

.vscode/launch.json
```diff
@@ -10,9 +10,8 @@
         "vendor/"
       ],
       "pathMappings": {
-        "/var/www/html/wp-content/plugins/stream-src": "${workspaceFolder}",
-        "/var/www/html/wp-content/plugins/stream": "${workspaceFolder}/build",
-        "/var/www/html": "${workspaceFolder}/local/public"
+        "/var/www/html/wp-content/plugins/stream": "${workspaceFolder}",
+        "/var/www/html": "${workspaceFolder}/local/public/wp"
```
bartoszgadomski, line 14: wp-env doesn't serve core from `local/public/wp` — that path is the Composer (`roots/wordpress`) install from `extra.wordpress-install-dir`. The container's `/var/www/html` is wp-env's own WordPress under `~/.wp-env/<hash>/WordPress`.

As written, Xdebug maps container core files onto a different WordPress copy, so stepping into core lands in files that only coincidentally match. The plugin mapping on line 13 is the one that matters here; I'd drop this entry (or point it at the wp-env directory).

.github/workflows/lint-and-test.yml
```diff
@@ -64,16 +57,16 @@ jobs:
       - name: Build
         run: npm run build
 
-      - name: Pull custom Docker images
-        run: docker compose pull wordpress
+      - name: Start wp-env
```
bartoszgadomski, line 60: Leftovers now that GHCR and Compose are gone:

- `WORDPRESS_IMAGE_VERSION: php8.2` (line 9) no longer has a consumer — PHP is pinned by `phpVersion` in `.wp-env.json`, and wp-env's env override is `WP_ENV_PHP_VERSION` (which `switch-to:php*` uses).
- `packages: read` (lines 19 and 73) was only needed for the `ghcr.io` login this PR removes.

Both are harmless, just dead config worth clearing while this file is already being touched.

bartoszgadomski (commented):

